PR #5, opened 2014-10-22
visual studio support

r-lyeh:
deprecated rand_r() calls removed();
log() disambiguation;
missing assert() header added;
portable endian.h header added (public domain license);


anthrotype:
:+1:

szabadka:
I will close this as these issues were addressed recently.

PR #17, opened 2014-11-20
Add Python bindings

khaledhosny:
This adds Python bindings (developed originally as https://github.com/khaledhosny/brotli-python). Everything is kept under `python` subdirectory, but distutils is very inflexible, so I had to symlink `dec` and `enc` directories from top level for the build to work (I wouldn’t insist on using a `setup.py` script myself, but it seems to be required for PyPI which I guess we would want to uploat to at some point). CC @behdad.


python/setup.py
```diff
@@ -0,0 +1,124 @@
...
+                        "dec/transform.h",
+                        "dec/types.h",
+                    ],
+                    language="c++",
+                    )
+
+setup(
+    name="Brotli",
+    version="0.1",
+    url="https://github.com/google/brotli"
```
anthrotype: @khaledhosny there's a missing comma at line 117.

apart from this, are there any plans to merge this pull request?

khaledhosny: Thanks, fixed.

khaledhosny:
FWIW, I need this for https://github.com/w3c/woff2-tests.

szabadka:
Hi Khaled. As per Google's open-source policy, you need to sign the [Google Individual Contributor License Agreement](https://developers.google.com/open-source/cla/individual?csw=1)(CLA), which you can do online.

khaledhosny:
I signed it now.

PR #24, opened 2015-02-15
visual studio support (again)

anthrotype:
this pull request is the same as @r-lyeh request #5 from October last year, with just some little modifications of mine.

There are three problems when trying to compile the C++ brotli encoder with Microsoft Visual C++ 12.0:

- "fast_log.h" is missing the assert header include:

    enc/fast_log.h(44) : error C3861: 'assert': identifier not found

- the use of function rand_r() inside "block_splitter.cc" makes the compiler raise "identifier not found". As @r-lyeh suggested, rand_r could be replaced by a combination of srand() and rand().

    enc/block_splitter.cc(100) : error C3861: 'rand_r': identifier not found

- in "write_bits.h", the attempt to import "endian.h" fails on Windows, which unfortunately does not provide this header. Besides, the include does not seem to be necessary here. In fact, the macro "IS_LITTLE_ENDIAN" which is used in write_bits.h is already specified inside "port.h" header, which is among the includes. 

    enc/write_bits.h(24) : fatal error C1083: Cannot open include file: 'endian.h': No such file or directory

As a workaround, @r-lyeh proposed a new portable "endian.h" header file. However, for the sake of detecting the byte order, I believe it would be sufficient to presume that when _WIN32 (or _WIN64) is defined, then the byte order will always little endian (ok, the XBox 360 has a big endian processor, but...).
So I added such Windows-specific #ifdef to "port.h", and removed @r-lyeh's "endian.h" file.

I would like to emphasize that it's only by using the patch by @r-lyeh (with or without the superfluous endian.h) that I was able compile the brotli Python extension on the Windows platform.

I have tested both gcc (via Cygwin or MinGW) and several Microsoft Visual C/C++ compilers.

Cygwin's gcc (POSIX) was able to compile only through @r-lyeh patch.

The MinGW-w64 gcc (targeting win32) was not able to compile even when using @r-lyeh's patch, so I gave up on gcc and tried one of MS Visual Studio compilers.

I first tried MS Visual Studio 2008 (the same used for Windows Python 2.7), but it was missing stdint.h header...

So I tried to force Python's distutils to use Visual Studio 2010. This time stdint.h was found, but the math.h module was missing the binary logarithm (log2) function... (of course, that could be rewritten as log(v)/log(2) or something).

Finally, I tried to make Python use MS Visual Studio 2013 as a compiler, and this time the Python extension compiled correctly, again thanks to @r-lyeh patch.

The brotli decoder (which is written in plain C) does compile successfully on Windows. It is the encoder, written in modern C++, which has the above mentioned issues. It would be nice if you could add Windows among the supported platforms. I'm interested because we are trying to use the brotli Python bindings in fontTools for decoding/encoding WOFF2.

Thanks a lot for your work.

Cosimo


anthrotype:
@khaledhosny 
in aac07290bee46acf5b9ee7821aeddf3b78f4ef70, I have slightly modified the setup.py script in order to avoid using symlinks to `../enc/` and `../dec/` directories. The reason is that Python for Windows (yes, again) cannot follow those symlinks and therefore the build fails.
In order to compile the Python extension on Windows, one can work around distutils' inability to use relative paths to the C/C++ source files by simply converting them to absolute paths.

@szabadka Apologies for my insistence, but I would like to ask you if Brotli's core developers are interested at all in making Brotli (specifically its encoder and Python bindings) compatible with the Windows platform? If not, I would have to find other strategies to be able to encode WOFF2 fonts using Python.

Thank you.

szabadka:
Sorry for the delay, I have not been actively working on brotli for a while. My plan now is to test and submit your non-python changes to our internal version and then update this github repository.

After that is done, for the remaining python part, you will have to sign a CLA on this page:
https://developers.google.com/open-source/cla/individual?csw=1

anthrotype:
@szabadka I signed the CLA on behalf of my employer (i.e., Dalton Maag Ltd).

szabadka:
I fixed the issues regarding the files in the enc directory.

Could you revert those files and try to compile it under visual studio? If it is still broken, please let me know.

What e-mail address and github user name did you use for the CLA?

anthrotype:
thanks for the fixes! I'm gonna try now and let you know the results.
For the email address we used `google.contributors [at] daltonmaag.com` (we created an ad-hoc Google Groups address for this purpose, as shown in the CLA web form).
My github username (i.e. @anthrotype) is linked to my work email address, `cosimo.lupo [at] daltonmaag.com`, and the latter email address is a member of the above mentioned google group.
Does that make sense?

anthrotype:
Yes, I can confirm the latest e60b7b846ab6a0a3de8d092b1a6aa1f1823dbb8c compile successfully under MS Visual C++ 12.0. 
Thank you very much for addressing this issue. 
I can close this pull request now.
I am going to open a new request, always Windows-specific, which has to do with the Python extension only.
Thanks again,

C.

PR #25, opened 2015-02-24
Make decoder code warning-free

irori:
The decoder is security critical module as it runs on browser against downloaded font data (potentially malicious). So let's keep it warning-free.


irori:
@szabadka
This is also needed to update Chromium's Brotli decoder, since Chromium builder treats warnings as errors.

PR #32, opened 2015-02-26
further fixes to allow compilation on MS Visual Studio 2010

anthrotype:
Hello,

I don't like playing the Windows guy here, but there are still some unresolved issues with Microsoft's C++ compiler.

The recent patches by @szabadka ensured the Brotli encoder compiles successfully when using MS Visual Studio 2013.
However, the problem is that the latest Python for Windows (versions 3.3 and 3.4) are compiled with Visual Studio 2010.
Now, from what I understood, there's a requirement that Python extensions be linked to the same C runtime library which Python itself uses:

https://docs.python.org/3.4/extending/windows.html

If that's the case, then we must define a `log2` (binary logarithm) function, which unfortunately is not present in `math.h` from Visual C++  2010.
Here's my attempt, which seemingly doesn't break anything else, since I only redefine log2 if `_MSC_VER <= 1600`.

Through this patch, the Python extension can be built on Windows using the same compiler version used to build CPython itself.
Well, at least for Python 3.3 and 3.4... previous versions are still built with Visual Studio 2008, but I still had no luck with that.

Cheers,

C.

### DIFF
--- a/enc/fast_log.h
+++ b/enc/fast_log.h
@@ -159,6 +159,13 @@ static const float kLog2Table[] = {
   7.9943534368588578f
 };
 
+#if defined(_MSC_VER) && _MSC_VER <= 1600
+// Calculates log2 of number (missing from MSC v.1600 required for Python bindings)
+inline double log2( double n ) {
+    return log( n ) / log( 2. );
+}
+#endif
+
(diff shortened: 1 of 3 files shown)


anthrotype:
#33 fixes this so I shall close it.
thank you!

PR #33, opened 2015-02-27
Fix encoder compilation error on MSVS 2010.

szabadka:
As reported by @anthrotype, log2() is missing from MSVS 2010.
This patch uses log() and a multiplication in FastLog2()
for _MSV_VER <= 1600 and uses FastLog2() in literal_cost.cc
instead of log2().

### DIFF
--- a/enc/fast_log.h
+++ b/enc/fast_log.h
@@ -164,7 +164,14 @@ static inline double FastLog2(int v) {
   if (v < (int)(sizeof(kLog2Table) / sizeof(kLog2Table[0]))) {
     return kLog2Table[v];
   }
+#if defined(_MSC_VER) && _MSC_VER <= 1600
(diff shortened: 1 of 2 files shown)


anthrotype:
:+1:

anthrotype:
unfortunately there is a problem in `literal_cost.cc:161`.
the `log2` function is used there without testing if the _MSC_VER <=1600.
so the build fails because of "identifier not found".

anthrotype:
if I change line 161 of literal_cost.cc to this, it works:

    -    float lit_cost = log2(static_cast<double>(in_window) / histo);
    +    float lit_cost = FastLog2(in_window) - FastLog2(histo);

szabadka:
I missed this one, sorry, fixed in #34

anthrotype:
you're fast! :)

PR #35, opened 2015-03-13
Makefile fixes and cleanups

mitya57:
- Distinguish between CC/CFLAGS, CPP/CPPFLAGS and CXX/CXXFLAGS.
  Do not store compiler flags in CPPFLAGS, which is for preprocessor.
- Use COMMON_FLAGS for flags that are for both C and C++.
- Drop -m64 flag which is wrong on 32-bit systems.

### DIFF
--- a/dec/Makefile
+++ b/dec/Makefile
@@ -2,7 +2,7 @@
 
 include ../shared.mk
 
-CPPFLAGS += -Wall
+CFLAGS += -Wall
 
 OBJS = bit_reader.o decode.o huffman.o safe_malloc.o streams.o
 
--- a/shared.mk
+++ b/shared.mk
@@ -1,19 +1,18 @@
 OS := $(shell uname)
 
-GFLAGS=-no-canonical-prefixes -fno-omit-frame-pointer -m64
-
-CPP = g++
-LFLAGS =
-CPPFLAGS = -c -std=c++0x $(GFLAGS)
+CC ?= gcc
+CXX ?= g++
 
 EMCC = emcc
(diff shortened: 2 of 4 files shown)


anthrotype:
hey @mitya57,
This patch gets rid of the this warning when compiling on OS X:

    -clang++: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated

Cool. However, why do you drop the `-m64` flag? Shouldn't that be set for 64-bit environments? 
thanks,

Cosimo

mitya57:
`-m64` is only needed when you are trying to compile 64-bit code on 32-bit systems. On 64-bit systems, it is default and is not needed.

mitya57:
In case I have not convinced you:

```
$ uname -m
x86_64
$ echo "int main() { return 0; }" > test.c
$ gcc test.c
$ file a.out
a.out: ELF 64-bit LSB  executable, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.24, not stripped
$ md5sum a.out 
102810d713e5daf3bb450e2dc96e225e  a.out
$ gcc -m64 test.c
$ md5sum a.out
102810d713e5daf3bb450e2dc96e225e  a.out
```

The `-m64` option makes no difference, resulting files are identical and 64-bit.

anthrotype:
well, you did convince _me_ :)
thank you for the clarification.

CoolOppo:
These are definitely needed changes, and I noticed all of these things myself when I was editing the `Makefile`.

PR #36, opened 2015-03-13
[python] Use BrotliDecompress()

khaledhosny:
So that we can use a callback to dynamically allocate the decompression
buffer, getting rid of the optional bufsize argument to
decompress.decompress().

### DIFF
--- a/python/brotlimodule.cc
+++ b/python/brotlimodule.cc
@@ -74,37 +74,39 @@ static PyObject* brotli_compress(PyObject *self, PyObject *args) {
   return ret;
 }
 
+int output_callback(void* data, const uint8_t* buf, size_t count) {
+  std::vector<uint8_t> *output = (std::vector<uint8_t> *)data;
+  output->insert(output->end(), buf, buf + count);
+  return (int)count;
+}
+
 PyDoc_STRVAR(decompress__doc__,
-"decompress(string[, bufsize]) -- Return decompressed string."
-"\n"
-"Optional arg bufsize is the initial output buffer size.");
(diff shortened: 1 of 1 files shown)


python/brotlimodule.cc
```diff
@@ -74,38 +74,39 @@ static PyObject* brotli_compress(PyObject *self, PyObject *args) {
   return ret;
 }
 
+int output_callback(void* data, const uint8_t* buf, size_t count) {
+  std::vector<uint8_t> *output = (std::vector<uint8_t> *)data;
+  output->insert(output->end(), buf, buf + count);
+}
```
szabadka: As anthrotype@ pointed out, please add a return statement.

khaledhosny: Fixed already.

python/brotlimodule.cc
```diff
@@ -74,38 +74,40 @@ static PyObject* brotli_compress(PyObject *self, PyObject *args) {
   return ret;
 }
 
+int output_callback(void* data, const uint8_t* buf, size_t count) {
+  std::vector<uint8_t> *output = (std::vector<uint8_t> *)data;
+  output->insert(output->end(), buf, buf + count);
+  return (int)count;
```
khaledhosny: @szabadka I’m just wondering if this cast is safe, shouldn’t the callback function return `size_t` to avoid this unnecessary cast?

szabadka: If should be safe because the decoder won't write huge chunks at once. See decode.c where we typically cast from int to size_t.

PR #43, opened 2015-03-23
[python] fix calling python script as subprocess on Windows

anthrotype:
Hello,
This patch fixes an issue which occurs when calling a python script as a subprocess from within another python script on Windows.
Instead of executing the child script through the shell, I run it with the same `sys.executable` that was used to run the parent script.
Besides, I'm using `subprocess.check_call` instead of `subprocess.call` to halt execution if subprocess exits with non-zero return code.
Cheers,

Cosimo

### DIFF
--- a/python/tests/compatibility_test.py
+++ b/python/tests/compatibility_test.py
@@ -2,7 +2,7 @@
 from __future__ import print_function
 import sys
 import os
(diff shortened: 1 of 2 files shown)


anthrotype:
thanks!

PR #65, opened 2015-04-03
Clean up some warnings in header files.

nemequ:
These warnings are triggered when compiling with -Wshadow or
-Wsign-compare.  Since they are in header files they will also be
triggered in any project attempting to use brotli while compiling
with those warnings enabled.

For an example of these being triggered in an external project, see https://travis-ci.org/quixdb/squash/jobs/56952895#L790


nemequ:
There are a *lot* more of these now (especially for `-Wsign-compare`), and since there doesn't seem to be any interest in merging it I don't really want to update the PR.

Issue #72, opened 2015-04-16
add brotli to PyPI repository


anthrotype:
It would be nice to add Brotli to the official Python Package Index, so that users can download it with a simple `pip install brotli`.

We could add just the sdist tarball, or also some pre-compiled wheel packages for Windows and Mac platforms, maybe built automatically via Travis and/or AppVeyor -- like [here](https://github.com/anthrotype/brotli-wheels)

/cc @khaledhosny

khaledhosny:
We probably need a release of some kind to by able to identify which version have been uploaded?

anthrotype:
@szabadka would you be ready to tag a release?

szabadka:
There will be some more improvements to the encoder in the coming weeks and I would prefer cutting a release after that.

anthrotype:
Ok, fair enough. Looking forward to those.
Thank you,
C.

szabadka:
I have one quick question about the tests in the python subdirectory. Is there any easy way to run them on Linux in a self-contained manner, i.e. without needing to install anything outside the root directory in the repository. Ideally the following would work:

$ python setup.py build_ext
$ python setup.py test

The first command succeeds, but when I run the second, I get 'ImportError: No module named brotli', which suggests to me that it is looking for brotli in a system-wide lib and not in the created build/ directory.

anthrotype:
you're right, it shouldn't do that.
I'll try to fix it, thanks.

anthrotype:
@szabadka this works: https://github.com/google/brotli/pull/80

szabadka:
Thanks, I confirmed that it works on Linux.

anthrotype:
Good! I can confirm it works on OSX and Windows too.

khaledhosny:
Before the release we would need also to make sure the Python module exposes all features of the brotli library (like the new encoder params).

CoolOppo:
I'm not one to really judge here, but if you guys are deciding on a versioning scheme, I take a look at [Semantic Versioning](http://semver.org)! This project would benefit from it over other versioning schemes because it makes it easy to determine when breaking changes have been made to the API, which is obviously a concern considering the last comment that @khaledhosny made in this thread.

anthrotype:
> Before the release we would need also to make sure the Python module exposes all features of the brotli library (like the new encoder params).

Yes, we should add the encoder's newly introduced `BrotliParams`.

What other feature would you like to expose in the Python module?
[...]

PR #105, opened 2015-05-08
[python] expose new encoder parameters as kwargs of brotli.compress

anthrotype:
I added support for the new BrotliParams to the Python extension, as requested in https://github.com/google/brotli/issues/72.
I also modified the bro.py script so that we can control these parameters from the command line.

I have a question. In the encode.h, it's written that the parameter "quality" has to be in the range from 0 to 11. I presume this range is inclusive. Now, if I set the quality between 1 and 11, the output changes accordingly. However, if I set it to 0, the output is always truncated to the same 10 bytes. Does this mean that 0 is an invalid value for the 'quality' parameter?

Thanks.

C.


python/bro.py
```diff
@@ -2,22 +2,26 @@
...
-__version__ = '0.1'
+__version__ = '1.0'
 
 
-BROTLI_MODES = {
-    'text': brotli.MODE_TEXT,
-    'font': brotli.MODE_FONT
+# default values of encoder parameters
+DEFAULT_PARAMS = {
+    'mode': brotli.MODE_TEXT,
```
szabadka: I just added a MODE_GENERIC compression mode, which is the default. This avoids the confusion that MODE_TEXT could equally mean 'no idea' and 'text for sure' before. Could you change your argument description and default accordingly?

anthrotype:
sure, I'll do that.

python/bro.py
```diff
@@ -49,27 +53,72 @@ def get_binary_stdio(stream):
...
+                        choices=list(range(16, 25)),
+                        help='Base 2 logarithm of the sliding window size. Range is '
+                        '16 to 24. Defaults to 22.')
+    params.add_argument('--lgblock', metavar="LGBLOCK", type=int,
+                        choices=[0] + list(range(16, 25)),
+                        help='Base 2 logarithm of the maximum input block size. '
+                        'Range is 16 to 24. If set to 0, the value will be set based '
+                        'on the quality. Defaults to 0.')
+    above9 = parser.add_argument_group(
+                        'encoder parameters respected only if quality > 9.')
```
szabadka: I think these low-level parameters should not be exposed to the binary. Their purpose is to find a balance between compression speed and compression density, just like the quality parameter. The reason I still have these here is that I don't have a quality 10 yet, which would be a slow but not extremely slow version. Once I have a reasonable compromise there,  I will probably remove these and rely solely on the quality (or maybe the mode as well), just like for qualities < 9.

anthrotype: I see. Which arguments shall I keep or remove precisely?

szabadka: Those under "// These settings will be respected only if quality > 9."
should be removed.

anthrotype: ok

anthrotype:
@szabadka OK, I've rebased on top of master, set MODE_GENERIC as default compression mode, and removed the additional low-level parameters from the command line tool's arguments (but I left them in the module's compress function)

szabadka:
As for quality 0, it is allowed, but currently is the same as quality 1. I found why it is broken now, I will fix it soon.

anthrotype:
thank you!

PR #122, opened 2015-06-26
Fix vs2010

anthrotype:
I get the following errors on Windows using Visual Studio 2010 when trying to compile the Python extension:

```
enc/backward_references.cc(26) : error C2124: divide or mod by zero
enc/backward_references.cc(140) : error C3861: 'log2': identifier not found
enc/backward_references.cc(145) : error C3861: 'log2': identifier not found
```

The patch tries to fix both issues.

For the `kInfinity` constant, I use `std::numeric_limits::infinity` as found in the `<limits>` header.

For the `log2` function (missing from Visual Studio 2010) which is used in `backward_references.cc`, I define a new `_log2` function inside `fast_log.h`. This MSVC specific workaround was previously incorporated in the `FastLog2` function, so I just extracted it from there.

We need to support Visual Studio 2010, otherwise it's not possible to compile the Python extension for Windows (Python 3.4 for Windows from Python.org is still built using Visual Studio 2010).

Cheers,

Cosimo

### DIFF
--- a/enc/backward_references.cc
+++ b/enc/backward_references.cc
@@ -17,13 +17,14 @@
 #include "./backward_references.h"
 
 #include <algorithm>
+#include <limits>
 #include <vector>
 
 #include "./command.h"
 
 namespace brotli {
 
-static const double kInfinity = 1.0 / 0.0;
+static const double kInfinity = std::numeric_limits<double>::infinity();
 
 // Histogram based cost model for zopflification.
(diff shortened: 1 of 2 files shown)


szabadka:
I fixed this in #123, could you test that it works now?

anthrotype:
@szabadka it works now, thanks! :+1:

PR #125, opened 2015-08-09
Let brotli compile with -Wunused-function.

nico:


eustas:
This function has been removed already.
If you still have problems with unused function warning, feel free to create a new pull request.

PR #133, opened 2015-08-10
fix compilation on Windows Python 2.7 + support for MINGW32 and Cygwin

anthrotype:
This includes:

- a patch to `setup.py` to force `distutils` use MS Visual C++ 10.0, from Visual Studio 2010, instead of the default 9.0, Visual Studio 2008, when compiling for **Windows Python 2.7**. The latter does not support many of the modern C++ features required to compile the Brotli encoder. Windows Python version 3 and above are compiled with 2010 so they are ok.

- two patches to `setup.py` to allow building Brotli with GCC for Windows, instead of the default Microsoft compiler. This can be done by passing the `--compiler=mingw32` option to setup.py.

- a patch to `streams.cc` to fix compilation under the Cygwin environment. Here GCC complains that 'malloc' and 'free' weren't declared in the current scope:

```
enc/streams.cc: In constructor ‘brotli::BrotliFileIn::BrotliFileIn(FILE*, size_t)’:
enc/streams.cc:92:32: error: ‘malloc’ was not declared in this scope
       buf_(malloc(max_read_size)),
                                ^
enc/streams.cc: In destructor ‘virtual brotli::BrotliFileIn::~BrotliFileIn()’:
enc/streams.cc:96:22: error: ‘free’ was not declared in this scope
   if (buf_) free(buf_);
                      ^
error: command 'gcc' failed with exit status 1
```

### DIFF
--- a/python/bro.py
+++ b/python/bro.py
@@ -9,9 +9,6 @@ import brotli
 import platform
 
 
-__version__ = '1.0'
-
-
 # default values of encoder parameters
 DEFAULT_PARAMS = {
     'mode': brotli.MODE_GENERIC,
@@ -54,7 +51,7 @@ def main():
     parser = argparse.ArgumentParser(
         prog='bro.py',
         description="Compression/decompression utility using the Brotli algorithm.")
-    parser.add_argument('--version', action='version', version='%(prog)s 1.0')
+    parser.add_argument('--version', action='version', version=brotli.__version__)
     parser.add_argument('-i', '--input', metavar='FILE', type=str, dest='infile',
                         help='Input file', default=None)
     parser.add_argument('-o', '--output', metavar='FILE', type=str, dest='outfile',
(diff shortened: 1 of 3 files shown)


anthrotype:
I just added a try/except clause to skip the `distutils.msvc9compiler` monkey patch when running setup.py under MinGW environment, whereby attempting to import the `msvc9compiler` module would raise a `DistutilsPlatformError`.

szabadka:
I created a separate PR for the header issues. Could you merge your python changes in one PR? After I merge that, we should be good to cut the release, right?

anthrotype:
thanks! sure, I'll do that now.

anthrotype:
OK, i have consolidated the two PR branches in a single one (the current one).

I have also added a `__version__` string constant to the brotli Python module, and set it to "0.1.0".
When installing the module, the setup.py uses regex to find the version string in the `brotlimodule.cc` file, so that we don't have to set it twice.

I compiled and run the tests on my local Windows and OS X machines.
All seems good now on the python side. You may wish to proceed cutting the release.

Thanks for the good work!

PR #136, opened 2015-08-11
support building compiled wheels packages using Appveyor + Travis 

anthrotype:
Here's my attempt at using CI systems to run automated builds of Brotli Python extension module.

You can see the latest build logs here:
https://travis-ci.org/anthrotype/brotli
https://ci.appveyor.com/project/anthrotype/brotli

Only Windows (via Appveyor) and OS X (via Travis) are supported here.
PyPI does not accept precompiled Linux wheels anyway: see http://pythonwheels.com/

On Windows, I test both 32 and 64-bit Python, 2.7.x and 3.4.x. The extension is compiled using MSVC 2010.

On OS X, I use the [MacPython/terrify](https://github.com/MacPython/terryfy) utilities in order to download and configure different python setups.

I could successfully compile Brotli on Apple's system Python (2.7.x), as well as on Homebrew Python 2.7.10 and 3.4.3. However, I was not able to build Brotli using official Python installers from Python.org. The latter are compiled using OS X 10.6 SDK and GCC 4.2, which don't support advanced C++0x features required by Brotli.

We could configure both `appveyor.yml` and `.travis.yml` to deploy the compiled wheels to some public wheelhouse (even directly upload to PyPI) whenever a new release is tagged. 

E.g. http://docs.travis-ci.com/user/deployment/pypi/

Let me know what you think.
Thanks

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "terryfy"]
+	path = terryfy
+	url = https://github.com/MacPython/terryfy.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language:
 env:
   matrix:
   - INSTALL_TYPE='system' VERSION=2.7
+  - INSTALL_TYPE='macpython' VERSION=2.7.10 CC=clang CXX=clang++
+  - INSTALL_TYPE='macpython' VERSION=3.4.3 CC=clang CXX=clang++
   - INSTALL_TYPE='homebrew' VERSION=2.7.10
   - INSTALL_TYPE='homebrew' VERSION=3.4.3
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+language:
+- objective-c
+env:
+  matrix:
+  - INSTALL_TYPE='system' VERSION=2.7
+  - INSTALL_TYPE='homebrew' VERSION=2.7.9
+  - INSTALL_TYPE='homebrew' VERSION=3.4.3
+install:
+- source terryfy/travis_tools.sh
+- get_python_environment $INSTALL_TYPE $VERSION venv
+- pip install --upgrade wheel
+script:
+- python setup.py build_ext test
+after_success:
+- pip wheel -w dist .
+before_deploy:
+- cd dist
+- "wheels=$(echo *.whl)"
+#deploy:
+# - TODO: upload the content of $wheels to a public wheelhouse
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language:
 env:
   matrix:
   - INSTALL_TYPE='system' VERSION=2.7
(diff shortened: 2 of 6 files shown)


anthrotype:
I eventually succeeded to compile the Brotli extension for Mac Python (i.e. the official installers from Python.org).

The problem was twofold:

1) Travis provides the old Apple GCC 4.2 under `/usr/local/bin/gcc-4.2`. When compiling the Brotli extension for Mac Python, distutils picks up `gcc-4.2` as compiler, because technically that was the one used to build the original Mac Python distribution. However, the problem is `gcc-4.2` does not support C++11 (it aborts because it does not recognise the command line option `-std=c++0x`). To work around that, I need to override `CC` and `CXX` variables in Travis respectively to `clang` and `clang++` -- the latter does support C++11.

2) Even after setting clang as compiler, two additional flags needs to be explicitly passed to it for the Mac Python build to succeed: namely `-stdlib=libc++` and `-mmacosx-version-min=10.7`. This is to make sure the clang's C++ standard library (with C++11 support) gets used instead of the earlier `libstdc++`. The second flag sets the minimum deployment target to 10.7 (the default for Mac Python would be 10.6), which is required for using `libc++`.

So, patch no. 1 is specific to Travis' setup and only applies to `.travis.yml`.
However, patch no. 2 must go in `setup.py`, since the two additional compiler flags need to be passed only when compiling C++ sources on the Mac platform with clang.

I think it's important that Brotli can successfully compiles on the official Python.org distributions, besides on the Apple built-in Python and homebrew one. Not just because it's the official Mac distribution, but also because the compiled wheels that are built for the latter will have the platform tag `macosx_10_6_intel`: that is, they will work on both 32 and 64 bit architectures and on any OS X version >= 10.6.

anthrotype:
If you want, I could open a separate pull request for the setup.py's patch 7e471ca, as it applies more generally, and keep the current PR only for Travis and Appveyor related stuff.

szabadka:
Do I understand it correctly that if we add these config files, then some third party sites will start cloning this repository and try to build it and run tests?

Is adding these files enough or do we have to do something else? Like register on travis and/or appveyor?

anthrotype:
I'm not sure how it works with organization accounts like google's.. For individuals, you simply login into Travis or Appveyor using one's own Github personal account, and then give the CI service access to the repositories you wish to build. After that, you can track them on `https://travis-ci.org/<username>/<project>` or `https://ci.appveyor.com/project/<username>/<project>`.
That's what I did with my brotli fork.
So, it is not automatic, you first need to grant Travis and Appveyor access to the repository.

szabadka:
OK, lets break this PR up in three parts, one for the setup.py fix, one for travis and one for appveyor. After merging the setup.py and travis PRs I will try registering google/brotli with travis. I saw a bunch of google projects there, so it should be ok. I am not sure about appveyor yet, I will ask around next week.

anthrotype:
OK, I'll do that.

PR #143, opened 2015-08-24
Fix all Microsoft Visual Studio bugs and errors and Add Microsoft Visual Studio Projectfiles

dominikhlbg:
- Fix: missing unistd.h
- Add: missing STDIN, STDOUT, STDERR constants
- Add: _CRT_SECURE_NO_WARNINGS constant to turn off error compiling
message "unsafe function or variable"
- Add: _CRT_NONSTDC_NO_DEPRECATE constant to turn off error compiling
message "deprecated postfix name"
- Fix: function OpenOutputFile write binary now


ericlaw1979:
It looks like the change may have introduced tab-based indentation in bro.cc while the rest of the file uses spaces?

ericlaw1979:
This change results in the output file having the `readonly` attribute set; see http://stackoverflow.com/a/11810168/126229 for an explanation of why that is.

szabadka:
Eric's version for opening the file seems more robust. Dominik, did you try that? If that would work for you, I will make these changes to our internal version and push them here.

dominikhlbg:
No fopen is more robust.
http://stackoverflow.com/questions/1658476/c-fopen-vs-open

open() works with flags "S_IREAD | S_IWRITE"
But is it useful to use readflags Windows: "S_IREAD" and UNIX: "S_IRUSR" when file is use writeonlyflag "O_WRONLY"?

ericlaw1979:
In Windows, if you fail to pass `S_IRUSR | S_IWUSR` to `open`, the resulting file is created with the NTFS `ReadOnly` attribute set so the resulting file cannot be edited (or deleted by code).

dominikhlbg:
You talk about <code>S_IRUSR | S_IWUSR</code>.
I talk about <code>S_IRUSR | S_IREAD</code>.

ericlaw1979:
Per MSDN, S_IWRITE also allows S_IREAD, so presumably the latter isn't truly necessary, although it is also harmless and a bit more clear about the intention.

gyf19:
+1

nitram509:
Hi,
I just stumbled upon this issue and would like to add my 2$cents.
I like to see MS Visual Studio related bugs fixed, as this makes the code more cross platform compatible. And I dislike to see the MS project file added. Simply because it is awkward to maintain two or more build files consistently.

As a side note:
Since a few days I have good and painless experience by using CMake for building "libbrotli".
As it already comes with numerous templates I was able to compile Java bindings with it. I was able to compile with all three: VS GUI, command line MS nmake (32bit & 64bit) and OSX make.
See https://github.com/nitram509/jbrotli for details.
Frankly said I'm not a CMake expert and I expect my version misses some settings ;-)

Thanks for listening, feedback is welcome.
Martin

eustas:
We hope to add MSVS project files generation soon. Sorry for very late response.

PR #163, opened 2015-09-23
create 'brot' command-line compression program

agordon:
'brot' supports the same common usage as gzip/bzip2/xz .

add two test scripts:
* brot-test.sh compares behaviour with gzip/bzip2/xz
* brot-size-test.sh compares compression sizes with gzip/bzip2/xz

(could be renamed to 'bro', but I didn't want to override the current 'bro' program).

### DIFF
--- a/tools/brot.cc
+++ b/tools/brot.cc
@@ -1,0 +1,393 @@
+/* Copyright (C) 2015 Assaf Gordon
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+   Compression/Decompression code copied from 'bro.cc':
+     Copyright 2014 Google Inc. All Rights Reserved.
+
+
+   a command-line version of 'brotli' compatible with the
+   common usage of gzip/bzip2/xz .
+*/
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <libgen.h>
+#include <unistd.h>
+#include <err.h>
+#include <string.h>
+#include <assert.h>
+#include <getopt.h>
+
+#include "../dec/decode.h"
+#include "../enc/encode.h"
+#include "../enc/streams.h"
+
+/* Program Version */
+#define VERSION "0.42"
+
+/* Email for Bug Reports */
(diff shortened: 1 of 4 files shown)


CoolOppo:
I feel like there's a better name that could be used. I do like `bro`, but if that's already in use, that isn't really an option.

khaledhosny:
I think the existing `pro` tool can be replaced by this one, behaving like existing tools is a big plus.

CoolOppo:
@khaledhosny I think you mean `bro`, and if so, I completely agree. The current command-line `bro` tool is extremely annoying to use. Both `gzip` and `xz` work excellently as command-line tools, and there's really no good reason to differ from those.

szabadka:
We decided to change the interface of the bro tool to match gzip.

Would you be willing to change bro.cc with these changes to the command line parsing? For easier review, it could be broken up into multiple pull requests, e.g. as a first step factor out the function that processes one file from main().

As for function naming and other style conventions, please see https://google-styleguide.googlecode.com/svn/trunk/cppguide.html

And in order to accept your contributions, you have to sign a CLA (see https://github.com/google/brotli/blob/master/CONTRIBUTING)

Arkanosis:
That would be great!

BTW shouldn't --best be an alias for -11 instead of -9? That will probably be a bit more complex to handle since -11 is not a short option anymore, though.

tools/brot.cc
```diff
@@ -0,0 +1,393 @@
...
+#include "../enc/streams.h"
+
+/* Program Version */
+#define VERSION "0.42"
+
+/* Email for Bug Reports */
+#define BUGREPORT "x@y.com"
+
+/* File extensions for brotli files */
+#define EXTENSION ".bro"
```
Kijewski: Was changed to `.br`.

tools/brot.cc
```diff
@@ -0,0 +1,393 @@
...
+#define EXTLEN    (strlen(EXTENSION))
```
Kijewski: `(sizeof(EXTENSION) - 1)`

tools/brot.cc
```diff
@@ -0,0 +1,393 @@
...
+        compression_quality = (ch - '9');
+        break;
+
+      case 'c':
+        opmode = COMPRESS;
+        to_stdout = 1;
+        break;
+
+      case 'd':
+        opmode = DECOMPRESS;
```
gvaish: May be allow only one of `c` or `d`.

Let the default value of `opmode` be `UNKNOWN` with:

```c
typedef struct _opmode {
   UNKNOWN,
   COMPRESS,
   DECOMPRESS
} opmode;
```

thedrow:
Any chance to get this merged?

zmichael:
Please Please Please merge it

agordon:
To anyone following this topic (the 'brot' stand-alone compression) program:
I've long ago signed google's CLA (in fact, on Oct 02, 2015 11:23 PDT).

I haven't been keeping track of brotli since I wrote this - but if anyone is interested in picking it up and making it work - they have my full support. If there are minor changes that I can help with - let me know.

eustas:
Soon I will be landing next brotli update. Among other things it contains new CLI...

PR #170, opened 2015-09-24
add -fPIC to make compatible with shared library

witchu:


eustas:
Hello.

shared.mk is used to extract common configuration from makefiles.

But currently we have no library targets...

What is the disadvantage of placing "-fPIC" simply to CFLAGS and CXXFLAGS environment variables?

witchu:
I made a lua binding, it must be a .so. It not user friendly to build my binding with
```
CFLAGS=-fPIC CXXFLAGS=-fPIC make
```
instead of just
```
make
```




Now, My makefile call
```
make $(MAKE) -C brotli/enc
make $(MAKE) -C brotli/dec
```
But I think I can embed `brotli/enc/*.o` and `brotli/dec/*.o` into my makefile.


Thanks

bagder:
I put up a separate build project that builds a shared and/or static libbrotli library out of the brotli source code and it should use -fPIC already when necessary: https://github.com/bagder/libbrotli

eustas:
Fixed (in premake5 generated makefiles)

PR #184, opened 2015-09-28
Support large inputs/outputs in memory adaptors

IIoTeP9HuY:
Hello!
I had a trouble using BrotliMem(In/Out) adaptors while compressing large chunk of data ( > 3 GB).
So I'm sending PR that is supposed to fix this problem.

### DIFF
--- a/enc/streams.cc
+++ b/enc/streams.cc
@@ -23,12 +23,12 @@
 
 namespace brotli {
 
-BrotliMemOut::BrotliMemOut(void* buf, int len)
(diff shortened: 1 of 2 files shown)


szabadka:
Thanks for the fix.

Before we can merge, you have to sign a CLA (see https://github.com/google/brotli/blob/master/CONTRIBUTING).

Thanks!

IIoTeP9HuY:
Sure! Just did it.

PR #187, opened 2015-10-01
Define the encoder dictionary in the .cc file and link only once.

szabadka:


anthrotype:
maybe you need to remove `enc/dictionary.h` and add `enc/dictionary.cc` to the list of sources and headers in the setup.py?

anthrotype:
never mind, the Appveyor build is passing now.

PR #193, opened 2015-10-02
Add tools/version.h that contains the brotli version.

szabadka:
This can be reused by all of the tools.
Set the version to the next git release tag.


joeyhoer:
Version is at 0.3.0 with no corresponding release?

szabadka:
0.3.0 will be the version of the next release. It has to be updated before cutting the release.

Issue #200, opened 2015-10-02
vector.data() not supported on MS VC++ 9.0


anthrotype:
`std::vector::data` is defined in C+11, so Microsoft Visual C++ 9.0 (aka Visual Studio 2008) does not support it and raises the following error:

```
python/brotlimodule.cc(194) : error C2039: 'data' : is not a member of 'std::vector<_Ty>'
        with
        [
            _Ty=uint8_t
        ]
```

The (uglier) method `&vector[0]` could be used instead of `vector.data()` to get the pointer to the underlying array (provided it isn't empty).

As you know, Python 2.7 for Windows is still compiled with Visual Studio 2008, so all extension modules should in theory be compiled using the same MSVC compiler version. After VS2008 was discontinued, Microsoft released an ad-hoc "Visual C++ Compiler for Python 2.7" (http://aka.ms/vcpython27) meant to be used for compiling extension for Windows Python 2.7.

In the current Brotli's `setup.py`, I added a patch to force distutils use Visual Studio 2010 for Python 2.7. This seems to work well so far, despite many people warn against mixing different C runtimes versions between the interpreter and the extension modules.

Since you recently added support for C++98, it'd be nice to also try support the old Microsoft compiler.

Thank you.

anthrotype:
thank you very much!

PR #201, opened 2015-10-02
replace vector.data() with &vector[0] for MSVC90 compatibility

anthrotype:
This patch is meant to address issue #200, by replacing all uses of `vector.data()` with `&vector[0]`.

I also removed the distutils patch inside setup.py, so that the correct MSVC compiler version is selected automatically based on the python version (9.0 for Py2.7, 10.0 for Python 3.4, etc.) 

The `import setuptools` at the beginning of setup.py is required to use the "MS Visual C++ for Python 2.7" (http://aka.ms/vcpython27). Unfortunately the stock distutils fails to locate the `vcvarsall.bat` file..
This issue is fixed in setuptools (which is now bundled along with pip with the latest Python.org installers, and is the recommended way to distribute Python packages).

I know it looks terrible -- but at least shows how with little change we can make it work even with MSVC 9.0 and Windows Python 2.7.

Let me know what you think.
Thanks

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,7 +18,7 @@ environment:
     - PYTHON: "C:\\Python27-x64"
       PYTHON_VERSION: "2.7.x"
       PYTHON_ARCH: "64"
-      WINDOWS_SDK_VERSION: "v7.1"
+      WINDOWS_SDK_VERSION: "v7.0"
 
     - PYTHON: "C:\\Python34-x64"
       PYTHON_VERSION: "3.4.x"
--- a/python/brotlimodule.cc
+++ b/python/brotlimodule.cc
@@ -191,7 +191,7 @@ static PyObject* brotli_decompress(PyObject *self, PyObject *args) {
 
   ok = BrotliDecompress(in, out);
   if (ok) {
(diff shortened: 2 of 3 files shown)


szabadka:
I fixed everything in enc/ in PR #206. Could you create a PR for the rest?

anthrotype:
Thanks!
I've just rebased and forced-push to this same PR branch.
 
There was still one remaining instance of `.data()` inside `python/brotlimodule.cc`:

```
$ git grep "\.data()"
python/brotlimodule.cc:    ret = PyBytes_FromStringAndSize((char*)output.data(), output.size());`
```

I changed it to the following:
```diff
if (ok) {
-    ret = PyBytes_FromStringAndSize((char*)&output[0], output.size());
+    ret = PyBytes_FromStringAndSize((char*)(output.size() ? &output[0] : NULL), output.size());
   } else {
```

Let me know if that is ok.

Let's cross fingers on the Appveyor Python 2.7 build: it should not be using its native MSVC 9.0...

anthrotype:
sorry for the typo: I meant, the Appveyor Py2.7 build should *now* be using MSVC90

anthrotype:
it's working! :)

PR #226, opened 2015-10-13
Binary search in Log2Floor

Nicoshev11:
[decode.c] use binary search in Log2Floor

Log2Floor implementation uses lineal search to find the floor of the log2, doing a binary search is slightly faster.

Because Log2Floor is only called from ReadHuffmanCode and the parameter passed is always less than 1024, l16_Log2Floor does a binary search in the lower 16 bits of the number, thus, in only 3 loops it gets the result

### DIFF
--- a/dec/decode.c
+++ b/dec/decode.c
@@ -331,6 +331,30 @@ static BROTLI_INLINE uint32_t Log2Floor(uint32_t x) {
   return result;
 }
 
+/* Calculates the Floor of Log2(x) using a binary search in the lower 16 bits,
+   Precondition: x<2**16 
+*/
+static BROTLI_INLINE uint32_t l16_Log2Floor(uint32_t x) {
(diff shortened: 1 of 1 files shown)


eustas:
Hello.

Thank you for your contribution.

We've already tried different approaches to calculate Log2Floor in decoder.
And decided to keep the simplest one, because:
* it requires less registers
* it use less instructions cache

But the most important reason is - it takes about 0.01% of decoding time.
So, optimizing it won't make any difference, but could cause performance problems in unexpected manner.

Thank you again for your interest to brotli.
Best regards,
  Eugene.

Nicoshev11:
Ok, and what functions are the main bottleneck of the decoding procces ? If you have some straightforward work to do, like porting C++ optimizations or functions to C so the decoder can use them please let me know,

Best regards,
Nico

PR #229, opened 2015-10-19
Fixed minor white-space formatting and ordering of elements

dsnet:
Fixed minor white-spacing issues that caused print-out to be slightly
confusing. Biggest change is in section 9.2, where an indent seemed
to indicate that some fields were part of the previous field, when
they were not related.

Also, changed the order that transforms are described in section 8
to match the enumeration values that are explicitly defined in
Appendix B.

### DIFF
--- a/docs/draft-alakuijala-brotli-07.nroff
+++ b/docs/draft-alakuijala-brotli-07.nroff
@@ -1243,20 +1243,20 @@ following C language function:
 
 .nf
    void InverseMoveToFrontTransform(uint8_t* v, int v_len) {
-     uint8_t mtf[256];
-     int i;
-     for (i = 0; i < 256; ++i) {
-       mtf[i] = (uint8_t)i;
-     }
-     for (i = 0; i < v_len; ++i) {
-       uint8_t index = v[i];
-       uint8_t value = mtf[index];
-       v[i] = value;
-       for (; index; --index) {
-         mtf[index] = mtf[index - 1];
-       }
-       mtf[0] = value;
-     }
+      uint8_t mtf[256];
+      int i;
+      for (i = 0; i < 256; ++i) {
+         mtf[i] = (uint8_t)i;
+      }
+      for (i = 0; i < v_len; ++i) {
+         uint8_t index = v[i];
+         uint8_t value = mtf[index];
+         v[i] = value;
+         for (; index; --index) {
+            mtf[index] = mtf[index - 1];
+         }
+         mtf[0] = value;
+      }
    }
 .fi
 
@@ -1795,7 +1795,7 @@ available in the brotli open-source project:
 https://github.com/google/brotli
 
 .ti 0
-15. Acknowledgements
+15. Acknowledgments
 
 The authors would like to thank Mark Adler for providing helpful review
 comments, validating the specification by writing an independent decompressor
--- a/docs/draft-alakuijala-brotli-07.nroff
+++ b/docs/draft-alakuijala-brotli-07.nroff
@@ -493,11 +493,11 @@ from most- to least-significant bit. The code lengths are
 initially in tree[I].Len; the codes are produced in tree[I].Code.
 
 .nf
-   1)  Count the number of codes for each code length. Let
-       bl_count[N] be the number of codes of length N, N >= 1.
+   1) Count the number of codes for each code length. Let
+      bl_count[N] be the number of codes of length N, N >= 1.
 
-   2)  Find the numerical value of the smallest code for each
-       code length:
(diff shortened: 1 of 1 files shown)


dsnet:
In case you're interested, I started a [third-party implementation of Brotli in Go](https://github.com/dsnet/compress/tree/brotli).

This implementation uses the RFC alone; as such I have avoided looking at the source code of Brotli.
This pull request was done based on things I noticed as I worked on my implementation using the RFC.

docs/draft-alakuijala-brotli-07.nroff
```diff
@@ -500,28 +500,28 @@ initially in tree[I].Len; the codes are produced in tree[I].Code.
        code length:
 
 .KS
-          code = 0;
-          bl_count[0] = 0;
-          for (bits = 1; bits <= MAX_BITS; bits++) {
-             code = (code + bl_count[bits-1]) << 1;
-             next_code[bits] = code;
-          }
+         code = 0;
```
szabadka: Here the indentation is actually correct, we use 3-space indentation by default.

docs/draft-alakuijala-brotli-07.nroff
```diff
@@ -500,28 +500,28 @@ initially in tree[I].Len; the codes are produced in tree[I].Code.
...
 
 .KS
-           for (n = 0;  n <= max_code; n++) {
-              len = tree[n].Len;
-              if (len != 0) {
-                 tree[n].Code = next_code[len];
-                 next_code[len]++;
-              }
-           }
+         for (n = 0;  n <= max_code; n++) {
```
szabadka: This should be indented one space more.

docs/draft-alakuijala-brotli-07.nroff
```diff
@@ -1434,52 +1434,52 @@ meta-block is the last one. The format of the meta-block header is
...
+                      as invalid)
+
+              0 - 7 bits: fill bits until the next byte boundary,
+                          must be all zeros
+
+              MSKIPLEN bytes of metadata, not part of the
+                      uncompressed data or the sliding window
+
+   MNIBBLES x 4 bits: MLEN - 1, where MLEN is the length
+            of the meta-block uncompressed data in bytes (if the
```
szabadka: This should be indented one space less, to match the previous paragraph on the same level.

docs/draft-alakuijala-brotli-07.nroff
```diff
@@ -1434,52 +1434,52 @@ meta-block is the last one. The format of the meta-block header is
...
+
+   1 bit:  ISUNCOMPRESSED, if set to 1, any bits of compressed
+           data up to the next byte boundary are ignored, and
+           the rest of the meta-block contains MLEN bytes of
+           literal data; this field is only present if the
+           ISLAST bit is not set (if the ignored bits are not
+           all zeros, the stream should be rejected as invalid)
 
    1-11 bits: NBLTYPESL, # of literal block types, encoded with
               the following variable length code (as it appears in
```
szabadka: I think these should be on the same level as above (i.e. 3 space less indentation).

docs/draft-alakuijala-brotli-07.nroff
dsnet: Done.

docs/draft-alakuijala-brotli-07.nroff
dsnet: Done.

docs/draft-alakuijala-brotli-07.nroff
dsnet: See previous comment.

docs/draft-alakuijala-brotli-07.nroff
dsnet: I could be wrong, but I count 9-spaces in my patch and 10-spaces in the original. The reason I removed the space was because I did notice that the spec used 3-space tabs consistently (for the most part).

https://github.com/google/brotli/blob/master/docs/draft-alakuijala-brotli-07.txt#L555-L574
You can see that the "2)" does not line up with the "3)".

szabadka: What I meant is that the indentation relative to the paragraph above ("Find the numerical value ...") should be 3 spaces (and likewise for the for loop below).

dsnet: I just removed the extra space for each paragraph. Now the number headings "1)", "2)", "3)", and the paragraph text, and the code are all aligned to 3-space tabs relative to the start of the line.

PR #236, opened 2015-10-23
Add .PHONY targets

maxnordlund:
To make `make` less confused.


eustas:
Fixed (with premake5 generated build).

PR #240, opened 2015-10-26
Added note about invalid distance values

ende76:
It might be helpful for implementers to mention briefly conditions that qualify an input stream for rejection. This has already been applied in other places, so I believe it would make sense for distances as well.

It is my understanding that a negative distance, or a zero distance should be rejected as invalid.

### DIFF
--- a/docs/draft-alakuijala-brotli-07.nroff
+++ b/docs/draft-alakuijala-brotli-07.nroff
@@ -865,6 +865,9 @@ Given a distance symbol "dcode" (>= 16 + NDIRECT), and extra bits
    distance = ((offset + dextra) << NPOSTFIX) + lcode + NDIRECT + 1
 .fi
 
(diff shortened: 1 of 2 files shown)


docs/draft-alakuijala-brotli-07.nroff
```diff
@@ -865,6 +865,9 @@ Given a distance symbol "dcode" (>= 16 + NDIRECT), and extra bits
    distance = ((offset + dextra) << NPOSTFIX) + lcode + NDIRECT + 1
 .fi
 
+If a decoded distance has a zero or negative value, the stream should
```
szabadka: Since only the first 16 distance symbols can resolve to zero or negative values, I think this paragraph should be moved up a bit, after line 843, something like:

"If a special distance symbol resolves to a zero or negative value, the stream should be rejected as invalid."

ende76:
Amended as suggested.

PR #242, opened 2015-10-26
Added note about invalid block type value in block switch commands

ende76:
Added an explicit note to reject a stream as invalid, if a block type in a block switch command is not within its 0..NBLTYPEx-1 range.
I believe it is of value to mention rejection conditions explicitly in the spec, and with a consistent phrase ("the stream should be rejected as invalid"), to make it easy for implementers to find these conditions through a simple text search.

### DIFF
--- a/docs/draft-alakuijala-brotli-07.nroff
+++ b/docs/draft-alakuijala-brotli-07.nroff
@@ -1008,9 +1008,10 @@ lengths and distances, respectively, is encoded in the meta-block
 header, and it must equal to the largest block type plus one in that
 block category. In other words, the set of literal, insert-and-copy
 length and distance block types must be [0..NBLTYPESL-1],
-[0..NBLTYPESI-1], and [0..NBLTYPESD-1], respectively. From this it
-follows that the alphabet size of literal, insert-and-copy length and
-distance block type codes is NBLTYPES + 2, NBLTYPESI + 2 and
+[0..NBLTYPESI-1], and [0..NBLTYPESD-1], respectively. If a block type
+has a value outside its respective range, the stream should be rejected as
+invalid. From this it follows that the alphabet size of literal, insert-and-copy
+length and distance block type codes is NBLTYPES + 2, NBLTYPESI + 2 and
 NBLTYPESD + 2, respectively.
 
 Each block count in the compressed data is represented with a pair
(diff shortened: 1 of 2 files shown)


docs/draft-alakuijala-brotli-07.nroff
```diff
@@ -1008,9 +1008,10 @@ lengths and distances, respectively, is encoded in the meta-block
 header, and it must equal to the largest block type plus one in that
 block category. In other words, the set of literal, insert-and-copy
 length and distance block types must be [0..NBLTYPESL-1],
-[0..NBLTYPESI-1], and [0..NBLTYPESD-1], respectively. From this it
-follows that the alphabet size of literal, insert-and-copy length and
-distance block type codes is NBLTYPES + 2, NBLTYPESI + 2 and
+[0..NBLTYPESI-1], and [0..NBLTYPESD-1], respectively. If a block type
+has a value outside its respective range, the stream should be rejected as
+invalid. From this it follows that the alphabet size of literal, insert-and-copy
+length and distance block type codes is NBLTYPES + 2, NBLTYPESI + 2 and
```
szabadka: I just noticed that NBLTYPES should be changed to NBLTYPESL here. Could you fix this as well?

szabadka: I am starting to wonder if we need this extra sentence here.

Since the block types are read using a prefix code with a certain alphabet size, there is no way we can decode a block type that it outside the range of its alphabet, right?

Was your intention to address the overflowing of the "previous block type +1" symbol? That behavior is already defined above:
"If the current block type is the maximal possible, then a block type symbol of 1 results in wrapping to a new block type of 0."

ende76: I had a whole thing written how it happened, because the stream can define its own symbols in a simple prefix tree.
When I now went back to the spec, I saw that it actually explicitly mentions:
> "If the integer value [for a symbol]
>   is greater than or equal to the alphabet size, then the stream
>   should be rejected as invalid."
I have missed that in my implementation thus far, and therefore it has been possible for a corrupt stream to create block type codes in a simple prefix code that exceed NBLTYPEx + 2.

I agree now that my suggested change is not needed. I will submit an amended change that fixes the type NBLTYPES -> NBLTYPESL.

ende76: This was probably not very clear. Hopefully better explanation: In my implementation, I was under the impression that a simple prefix code can define arbitrary symbols, as long as they are encoded with ALPHABET_SIZE bits. However, the spec clearly mentions that the integer value of a symbol must be within [0..ALPHABET_SIZE-1].

I had missed that before, thus my implementation was wrong. That allowed a corrupt stream to inject invalid block types in a prefix code. I _thought_ the underlying cause was in not rejecting invalid block types, but the real cause was in not rejecting invalid symbols in simple prefix codes.

PR #250, opened 2015-11-02
Suggested RFC changes

dsnet:


dsnet:
This is minor, but [the example given](https://github.com/google/brotli/blob/master/docs/draft-alakuijala-brotli-07.txt#L847-L851) for when only one single non-zero code length is provided, uses symbol 16. But practically, speaking, wouldn't that still require the extra 2-bits that follow a symbol 16 (even if the symbol itself is 0 bits)? Wouldn't it make more sense for the single symbol to just be symbol 8 (of 0 bits)? The subsequent paragraph dictating that the sum must be 32768 implies that the 0-bit symbol 8 could fill up all 256 symbols of a literal tree, since we can stop "reading" these 0-bit symbols once the tree is filled up. Why waste space encoded the extra bits of symbol 16?

docs/draft-alakuijala-brotli-07.nroff
```diff
@@ -1745,13 +1774,17 @@ The decoding algorithm that produces the uncompressed data is as follows:
...
             compute context ID, CIDD from CLEN
-            read distance code using HTREED[CMAPD[4 * BTYPE_D + CIDD]]
-            compute distance by distance short code substitution
-         move backwards distance bytes in the uncompressed data and
-            copy CLEN bytes from this position to the uncompressed
-            stream, or look up the static dictionary word, transform
-            the word as directed, and copy the result to the
-            uncompressed stream
+            read distance code using HTREED[CMAPD[4*BTYPE_D + CIDD]]
+            if distance code is not zero,
```
szabadka: It seems you accidentally removed a line from here:

"compute distance by distance short code substitution"

dsnet:
:+1: That's all the changes I got. Thanks for reviewing all of them!

I certainly hope Brotli takes off; the web could really use a new compression format.

PR #255, opened 2015-11-06
FIX: Typo in reference to NBLTYPESL, Minor: added missing word _lengths_ to insert-and-copy lengths

ende76:
Looks like the word "lengths" was missing.
Note: the line is now 73 characters long. There are a couple of lines as long as that one. It might have been a deliberate choice to omit the word to keep the line short.
If so, please reject.

### DIFF
--- a/docs/draft-alakuijala-brotli-07.nroff
+++ b/docs/draft-alakuijala-brotli-07.nroff
@@ -1741,9 +1741,9 @@ The decoding algorithm that produces the uncompressed data is as follows:
          read distance context map, CMAPD[]
       else
          fill CMAPD[] with zeros
-      read array of prefix codes for literals, HTREEL[]
-      read array of prefix codes for insert-and-copy lengths, HTREEI[]
-      read array of prefix codes for distances, HTREED[]
+      read array of literals prefix codes, HTREEL[]
+      read array of insert-and-copy lengths prefix codes, HTREEI[]
+      read array of distances prefix codes, HTREED[]
       do
          if BLEN_I is zero
             read block type using HTREE_BTYPE_I and set BTYPE_I
--- a/docs/draft-alakuijala-brotli-07.nroff
+++ b/docs/draft-alakuijala-brotli-07.nroff
@@ -1741,9 +1741,9 @@ The decoding algorithm that produces the uncompressed data is as follows:
          read distance context map, CMAPD[]
       else
          fill CMAPD[] with zeros
-      read array of literals prefix codes, HTREEL[]
-      read array of insert-and-copy lengths prefix codes, HTREEI[]
-      read array of distances prefix codes, HTREED[]
+      read array of literal prefix codes, HTREEL[]
+      read array of insert-and-copy length prefix codes, HTREEI[]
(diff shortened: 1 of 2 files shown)


docs/draft-alakuijala-brotli-07.nroff
```diff
@@ -1646,11 +1646,11 @@ commands. Each command has the following format:
       Insert length number of literals, with the following format:
 
          Block type code for next literal block type, appears
-            only if NBLTYPESI >= 2 and the previous literal
+            only if NBLTYPESL >= 2 and the previous literal
```
szabadka: Thanks for catching this!

docs/draft-alakuijala-brotli-07.nroff
```diff
@@ -1742,7 +1742,7 @@ The decoding algorithm that produces the uncompressed data is as follows:
       else
          fill CMAPD[] with zeros
       read array of prefix codes for literals, HTREEL[]
-      read array of prefix codes for insert-and-copy, HTREEI[]
+      read array of prefix codes for insert-and-copy lengths, HTREEI[]
```
szabadka: Since this is in a .nf .fi block, we have to make sure that the line length does not exceed 72. 

We can make it fit like this:
         read array of literal prefix codes, HTREEL[]
         read array of inset-and-copy length prefix codes, HTREEI[]
         read array of distance prefix codes, HTREED[]

szabadka: (I just noticed I had a typo there: inset->insert)

docs/draft-alakuijala-brotli-07.nroff
dsnet: :+1: Yep! This is my fault :/

ende76: @dsnet Mistakes happen. It's very hard to write something like this, where potentially every single word carries actual information. And then it gets even harder, if the goal is to convey the correct information, but in a way that's easy to understand.

For what it's worth, I found all your changes very helpful, and had actually written a fairly long comment on why I found it necessary to mention that duplicate symbols in simple prefix codes are to be considered invalid, when I then noticed that in the newest version, you had already addressed it.

It's a process, and I believe it's going well overall. ;)

docs/draft-alakuijala-brotli-07.txt
```diff
@@ -1854,9 +1854,9 @@ Internet-Draft                   Brotli                     October 2015
 
          else
             fill CMAPD[] with zeros
-         read array of prefix codes for literals, HTREEL[]
-         read array of prefix codes for insert-and-copy, HTREEI[]
-         read array of prefix codes for distances, HTREED[]
+         read array of literals prefix codes, HTREEL[]
```
szabadka: One more nit: I think it should be 'literal prefix codes'. No need to put both 'literal' and 'code' in plural. (and similarly for the other two).

ende76: I fully agree. That was simply an oversight on my part. Give me a minute.

PR #258, opened 2015-11-09
Implementation of paralell buffer compression based on std::thread

erikzenker:
* Added new param n_threads to brotli params
* brolti::BrotliCompressBufferParallel runs n_threads to compress input data

### DIFF
--- a/enc/encode.h
+++ b/enc/encode.h
@@ -39,6 +39,7 @@ struct BrotliParams {
         quality(11),
         lgwin(22),
         lgblock(0),
+        n_threads(1),
         enable_dictionary(true),
         enable_transforms(false),
         greedy_block_split(false),
@@ -63,6 +64,8 @@ struct BrotliParams {
   // Base 2 logarithm of the maximum input block size. Range is 16 to 24.
   // If set to 0, the value will be set based on the quality.
   int lgblock;
+  // Number of threads used for parallel compression.
+  int n_threads;
 
   // These settings are deprecated and will be ignored.
   // All speed vs. size compromises are controlled by the quality param.
--- a/enc/encode_parallel.cc
+++ b/enc/encode_parallel.cc
@@ -18,6 +18,9 @@
 
 #include <algorithm>
 #include <limits>
+#include <iostream>
+#include <thread>
+#include <utility>
 
 #include "./backward_references.h"
 #include "./bit_cost.h"
@@ -216,6 +219,28 @@ bool WriteMetaBlockParallel(const BrotliParams& params,
 
 }  // namespace
 
+std::pair<size_t, size_t>  consecutiveMapping(const size_t tid, const size_t nThreads, const size_t nElements){
+
+    const size_t elementsPerThread = static_cast<size_t>(ceil(static_cast<double>(nElements) / nThreads));
+    const size_t min_i             = tid * elementsPerThread;
+    const size_t max_i             = std::min(min_i + elementsPerThread, nElements);
+
+
+    if(tid > nElements - 1){
+	return std::make_pair(0,0);
+    }
+	
+    if(min_i > nElements){
+	return std::make_pair(0,0);
+    }
+
+	
+    return std::make_pair(min_i, max_i);
+	
+}
+
+    
+    
 int BrotliCompressBufferParallel(BrotliParams params,
                                  size_t input_size,
                                  const uint8_t* input_buffer,
@@ -236,6 +261,7 @@ int BrotliCompressBufferParallel(BrotliParams params,
   } else if (params.lgwin > kMaxWindowBits) {
     params.lgwin = kMaxWindowBits;
   }
+  
   if (params.lgblock == 0) {
     params.lgblock = 16;
     if (params.quality >= 9 && params.lgwin > params.lgblock) {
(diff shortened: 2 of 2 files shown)


psychocoderHPC:
**I am no project maintainer**
I reviewed your pull request and will give you some feedback.

enc/encode.h
```diff
@@ -39,6 +39,7 @@ struct BrotliParams {
         quality(11),
         lgwin(22),
         lgblock(0),
+	n_threads(1),
```
psychocoderHPC: please use `spaces` instead of a `tab`

enc/encode_parallel.cc
```diff
@@ -216,6 +219,28 @@ bool WriteMetaBlockParallel(const BrotliParams& params,
 
 }  // namespace
 
+std::pair<size_t, size_t>  consecutiveMapping(const size_t tid, const size_t nThreads, const size_t nElements){
+
+    const size_t elementsPerThread = ceil((float) nElements / nThreads);
```
psychocoderHPC: should be `ceil(static_cast<double>(nElements) / nThreads);` to avoid float overflows

enc/encode_parallel.cc
```diff
@@ -246,43 +272,83 @@ int BrotliCompressBufferParallel(BrotliParams params,
...
+	      size_t min_block_i    = 0;
+	      size_t max_block_i    = 0;
+	      std::tie(min_block_i, max_block_i) = consecutiveMapping(thread_i, params.n_threads, max_num_blocks);
+  
+	      for(size_t block_i = min_block_i; block_i < max_block_i; ++block_i){
+		  //std::cout << "[" <<  thread_i << "/"<< params.n_threads << "/" << block_i <<"] Compress block " << std::endl;
+      
+		  size_t input_block_size = std::min(max_input_block_size, input_size - ((block_i+1) * max_input_block_size));
+		  size_t pos              = block_i * max_input_block_size;
+		  out_sizes[block_i]      = 1.2 * input_block_size + 1024;
```
psychocoderHPC: You mixed `double`, `size_t` and `int` operation.
The original version was `size_t out_size = input_block_size + (input_block_size >> 3) + 1024` without a magic `1.2`

erikzenker:
Thx you for the review, I updated my pull request.

ende76:
@erikzenker Not a maintainer either, but I'm fairly familiar with the Brotli specification, albeit from a _decoder_ perspective rather than an encoder.

I am surprised that it should be possible to parallelize metablock encoding. It is possible that my understanding is incorrect. I will try to explain why I think that parallel encoding can't work with Brotli.

In Brotli, there are three mutable concepts that get initialized at the start of the stream, and then live across metablock boundaries. They have an impact on the decoding behavior of each block, and they also potantially get mutated within each block.

The three concepts are:
1. The output window (size defined by WSIZE in the stream header), initialized as empty. Used for back references, can span multiple blocks.
2. The literals buffer of the last two literals, referred to as p1 and p2 in the specification, initialized as p1 = 0, p2 = 0. Used to calculate Context ID for literals.
3. The distances (ring) buffer of the last four distances, initalized as [4, 11, 15, 16], where 4 is the most recent distance. Used in the decoding of distances, where distance codes can refer to these values.

As far as I understand it, parallel decoding of blocks is not possible because of those concepts, since subsequent bytes may depend on the state of any of these three concepts, which might have been mutated by actions on earlier compressed data.

Unfortunately, all of the test cases that actually contain multiple metablocks are structured in a way that doesn't test for this. I.e. in all of those cases, all of the output data is really encoded in just one of the metablocks. Because of that, the three stream-wide concepts above do not actually get mutated in those test cases, and thus your parallel implementation passes, when really, it should conceptually fail.

@szabadka Am I overlooking something?

szabadka:
On Tue, Nov 10, 2015 at 2:28 PM, Thomas Pickert <notifications@github.com>
wrote:

> @erikzenker <https://github.com/erikzenker> Not a maintainer either, but
> I'm fairly familiar with the Brotli specification, albeit from a *decoder*
> perspective rather than an encoder.
>
> I am surprised that it should be possible to parallelize metablock
> encoding. It is possible that my understanding is incorrect. I will try to
> explain why I think that parallel encoding can't work with Brotli.
>
> In Brotli, there are three mutable concepts that get initialized at the
> start of the stream, and then live across metablock boundaries. They have
> an impact on the decoding behavior of each block, and they also potantially
> get mutated within each block.
>
> The three concepts are:
> 1. The output window (size defined by WSIZE in the stream header),
> initialized as empty. Used for back references, can span multiple blocks.
> 2. The literals buffer of the last two literals, referred to as p1 and p2
> in the specification, initialized as p1 = 0, p2 = 0. Used to calculate
> Context ID for literals.
> 3. The distances (ring) buffer of the last four distances, initalized as
> [4, 11, 15, 16], where 4 is the most recent distance. Used in the decoding
> of distances, where distance codes can refer to these values.
>
> As far as I understand it, parallel decoding of blocks is not possible
> because of those concepts, since subsequent bytes may depend on the state
> of any of these three concepts, which might have been mutated by actions on
> earlier compressed data.
>
> Unfortunately, all of the test cases that actually contain multiple
> metablocks are structured in a way that doesn't test for this. I.e. in all
> of those cases, all of the output data is really encoded in just one of the
> metablocks. Because of that, the three stream-wide concepts above do not
> actually get mutated in those test cases, and thus your parallel
> implementation passes, when really, it should conceptually fail.
>
> @szabadka <https://github.com/szabadka> Am I overlooking something?
>
Each thread of the parallel implementation can have access to the whole
input data, therefore each of them has access to 1. the window (which is
just past input data) and 2. to the last two bytes of the previous input
block.

To address 3. the parallel encoder has to ensure that it will not use the
last distances to create a distance short code for the first two commands
of the meta-block (it is the choice of the encoder whether to make use of
the distance ringbuffer of to just output raw distances). After the first
two commands, the distance ringbuffer is fully initialized based on the
current meta-block and the encoder can start using it.

So yes, it is possible to do parallel metablock encoding.

ende76:
Ah, right, I did not consider that what I (as a decoder) typically consider "output" actually becomes "input" for an encoder.

I apologize for the confusion!

erikzenker:
Are there some test for the parallel encoding interface preset ? 
These tests could be used to verify the correctness of this
parallel code.

eustas:
Hello.

 encode_parallel is mostly a skeleton / proof of concept. Currently we have no plans to develop it.

PR #261, opened 2015-11-11
Fixed issue when linking with C++/CLI application

leidegre:
The Microsoft Visual C++ compiler, or specifically linker, generates an error about a possible uninitialized pointer when linking with a C++/CLI application if `output` is left uninitialized. This fixes that issue.

I get that this is a really minor thing but at least this way no one else needs to run into this issue if they want to use brotli in a C++/CLI project.

### DIFF
--- a/enc/encode.cc
+++ b/enc/encode.cc
@@ -629,7 +629,7 @@ int BrotliCompressWithCustomDictionary(size_t dictsize, const uint8_t* dict,
                                        BrotliIn* in, BrotliOut* out) {
   size_t in_bytes = 0;
   size_t out_bytes = 0;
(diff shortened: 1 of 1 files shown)


eustas:
Hello.

 Thank you for your patch.
 Unfortunately, we can't merge it until you sign CLA. More details here:
https://github.com/google/brotli/blob/master/CONTRIBUTING.md

Best regards,
  Eugene.

leidegre:
OK, an CLA has been submitted. Pending review...

eustas:
BrotliCompressWithCustomDictionary have been moved to compressor.cc and refactored. (And soon will be deprecated)

Issue #275, opened 2015-12-02
WriteMetadata may write outside the buffer provided to it


nemequ:
Picked this up running the Squash unit tests in AddressSanitizer.  Basically, the plugin gives brotli 6 bytes to write to, but the write is 8 bytes at a time.  See quixdb/squash#168 for background, info from ASan follows:

```
==9544==ERROR: AddressSanitizer: unknown-crash on address 0x611000005389 at pc 0x7f45a295fd04 bp 0x7ffe63ec49f0 sp 0x7ffe63ec49e0
WRITE of size 8 at 0x611000005389 thread T0
    #0 0x7f45a295fd03 in BROTLI_UNALIGNED_STORE64(void*, unsigned long) /home/nemequ/local/src/squash/plugins/brotli/brotli/enc/././././port.h:148
    #1 0x7f45a295fd03 in brotli::WriteBits(int, unsigned long, int*, unsigned char*) /home/nemequ/local/src/squash/plugins/brotli/brotli/enc/./write_bits.h:63
    #2 0x7f45a295fd03 in brotli::BrotliCompressor::WriteMetadata(unsigned long, unsigned char const*, bool, unsigned long*, unsigned char*) /home/nemequ/local/src/squash/plugins/brotli/brotli/enc/encode.cc:545
    #3 0x7f45a28f9c92 in squash_brotli_compress_stream /home/nemequ/local/src/squash/plugins/brotli/squash-brotli.cpp:191
    #4 0x7f45a28f9c92 in squash_brotli_process_stream /home/nemequ/local/src/squash/plugins/brotli/squash-brotli.cpp:261
    #5 0x7f45a6f344fe in squash_stream_process_internal /home/nemequ/local/src/squash/squash/stream.c:699
    #6 0x401ad7 in flush_test /home/nemequ/local/src/squash/tests/flush.c:23
    #7 0x7f45a69ff45a  (/lib64/libglib-2.0.so.0+0x6f45a)
    #8 0x7f45a69ff622  (/lib64/libglib-2.0.so.0+0x6f622)
    #9 0x7f45a69ff622  (/lib64/libglib-2.0.so.0+0x6f622)
    #10 0x7f45a69ff82d in g_test_run_suite (/lib64/libglib-2.0.so.0+0x6f82d)
    #11 0x7f45a69ff850 in g_test_run (/lib64/libglib-2.0.so.0+0x6f850)
[...]

PR #277, opened 2015-12-04
Fix WriteMetadata (unaligned and out-of-bounds write).

eustas:


eustas:
https://github.com/google/brotli/issues/275

PR #278, opened 2015-12-08
don't pass `-std=c++0x` option to gcc in setup.py

anthrotype:
Now that Brotli can be built with older compilers that don't support C++11, I think it's fine to turn that off in the `setup.py` script which compiles the Python extension module.

### DIFF
--- a/python/brotlimodule.cc
+++ b/python/brotlimodule.cc
@@ -21,9 +21,9 @@ static int mode_convertor(PyObject *o, BrotliParams::Mode *mode) {
   }
 
   *mode = (BrotliParams::Mode) PyInt_AsLong(o);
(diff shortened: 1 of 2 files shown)


anthrotype:
At [behdad/fonttools](https://github.com/behdad/fonttools/issues/429) project, we would like to switch to Travis CI container-based infrastructure (ie. `sudo: false`), however we also require to install gcc-4.8 (via `sudo apt-get`) since the Travis' built-in gcc version does not support C++11.

Through this patch, we can compile the Brotli extension using the old gcc, so no more need to sudo and we can use the new infrastructure.

/ping @szabadka

anthrotype:
thanks!


--
Cosimo

PR #304, opened 2016-01-25
Fix compilation warnings in brotlimodule.cc

eustas:


python/brotlimodule.cc
```diff
@@ -14,13 +14,27 @@ using namespace brotli;
 
 static PyObject *BrotliError;
 
+static int as_uint8(PyObject *o, int* result) {
```
szabadka: The name of this function is a bit confusing as there are no uint8 data types anywhere. Maybe name it as_bounded_int() and add the two bounds as parameters. Then most of the code below could be simplified, e.g. 'if (!as_bounded_int(o, lgwin, 10, 24)) { ...

Issue #307, opened 2016-01-28
BROTLI_BUILD_PORTABLE has dangerous default setting


ncopa:
firefox 44.0 recently shipped with brotli support. On Alpine Linux (which uses musl libc) it [crashed](https://bugzilla.mozilla.org/show_bug.cgi?id=1243724) due to it was not built with BROTLI_BUILD_PORTABLE set. (or BROTLI_SAFE_MEMMOVE to be sepcific)

There was no warning and there was no way to catch it til after the fact that it had happened. (it wouldn't been caught if we wouldn't build with FORTIFY_SOURCE).

I would recommend that you flip the default setting around and by default build with `BROTLI_BUILD_PORTABLE` or change it to `BROTLI_BUILD_NONPORTABLE`, because the memcpy implementation is outside of your control, you can not verify or warn if things gets miscompiled and relying on "undefined behavior" is generally a bad idea.

eustas:
Going to merge is ASAP.

PR #308, opened 2016-01-28
Make memmove16 safe.

eustas:


eustas:
This will fix Issue #307

Issue #174, opened 2015-09-25
Incompilable with Intel, few tweaks and options miissing


Sanmayce:
Hi Brotli team,
despite me being one of Google "haters" let me share my 2 cents on current Brotli status.

In next several months I intend to juxtapose several high-performance textual compressors with one goal in mind - showing most balanced ones for high-ratio/high-decompression-speed scenario.

Yesterday I downloaded your 'master' zip and compiled (with several syntactic changes) with Intel v15.0 optimizer.

In my incoming showdown I want to include Brotli wanting to see how it performs in its best environment, I speak textual (mostly English texts) files.

[Question #1:]
Since my goal is to show tightness&decompression-speed top-performers, are following enforced defaults best?

struct BrotliParams {
  BrotliParams()
//      : mode(MODE_GENERIC),
//        quality(11),
//        lgwin(22),
//        lgblock(0),
//        enable_dictionary(true),
//        enable_transforms(false),
//        greedy_block_split(false),
//        enable_context_modeling(true) {}

      : mode(MODE_TEXT),
        quality(11),
        lgwin(24),
        lgblock(24),
        enable_dictionary(true),
        enable_transforms(false),
        greedy_block_split(false),
        enable_context_modeling(true) {}

It would be very good to make these command line toggleable, no?

[Question #2:]
Your little announcement makes the impression Brotli is something special on text, what do I miss to see that? My quick test shows goodness but not greatness?
The below stats are for your yesterday commit compiled with Intel v15.0 (/O3 used), Brotli outperforms Shifune, but in decompression-speed department 3x is no joke, don't tell me if I use a browser or some English texts full-text browser/searcher Brotli will load 'dickens' faster than Zstd or even Shifune.

D:\>bro_Intel15.exe -i dickens -o dickens.brotli -v
Brotli compression speed: 0.200944 MB/s

D:\>bro_Intel15.exe -i dickens.brotli -o dickens -v -d -f
Brotli decompression speed: 142.945 MB/s

D:\>bro_Intel15.exe -i dickens.brotli -o dickens -v -d -f
Brotli decompression speed: 138.861 MB/s

D:\>bro_Intel15.exe -i dickens.brotli -o dickens -v -d -f
Brotli decompression speed: 145.079 MB/s

D:\>bro_Intel15.exe -i dickens.brotli -o dickens -v -d -f -r 5
Brotli decompression speed: 144.647 MB/s

[...]

PR #311, opened 2016-02-05
Make stdin/stdout work on Windows for tools/bro

fasterthanlime:
The code in master works well on Linux/OSX, not so well on Windows. This PR makes it work for all 3 main platforms. Props to #174 for giving helpful hints!

Tested on OSX 10.11.3 with clang 7.0.2, and Windows 10 with msys2, gcc 5.2.0


eustas:
Very sorry for that big latency (encoder was under construction). Fixed.

PR #313, opened 2016-02-11
brotlidump.py: disassemble brotli file

jneb:
This program is for educational and debug purposes and does nothing more (or less)  than dump the contents of brotli file, in an attempt to figure out the documentation.
I hope you find it useful.

### DIFF
--- a/python/brotlidump.py
+++ b/python/brotlidump.py
@@ -1,0 +1,1121 @@
+#!python3
+"""Program to dump contents of brotli files showing the compression format.
+"""
+import struct
+from operator import itemgetter
+from itertools import accumulate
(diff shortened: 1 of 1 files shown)


anthrotype:
That's cool!

jneb:
Oh, yes, a few remarks:

- It is Python3
- It is under construction, but most of the metablock header fields are supported
- I found with it that the sample compressed alice file uses signed integer context mode (??)

jneb:
Moved to the other pull request, in the "dumper" branch, so I can keep the code updated.
See the pull request with (revisited) in the title.
Sorry about the confusion

PR #317, opened 2016-02-19
Fix win32 fopen parameters

eustas:


eustas:
Many thanks to Matt Mahoney for discovery and to jibz for the fix (http://encode.ru/threads/2313-Brotli?p=46641&viewfull=1#post46641)

PR #321, opened 2016-02-23
Add custom dictionary feature binding

eustas:


python/brotlimodule.cc
```diff
@@ -148,8 +154,17 @@ static PyObject* brotli_compress(PyObject *self, PyObject *args, PyObject *keywd
   if (lgblock != -1)
     params.lgblock = lgblock;
 
-  ok = BrotliCompressBuffer(params, length, input,
-                            &output_length, output);
+  if (custom_dictionary_length == 0) {
```
szabadka: There should be a check to make sure that dictionary is shorter than the window, and set the error string to something appropriate otherwise.

eustas:
Fixed doc / added custom dictionary truncation.

szabadka:
LGTM

PR #331, opened 2016-03-19
Fix build with -Wmissing-declarations.

PiotrSikora:
While there, add -Wmissing-prototypes and -Wmissing-declarations
to shared.mk in order to catch similar errors in the future.


eustas:
Thank you!

Issue #326, opened 2016-03-08
libbrotli


bagder:
Hey,

I created the sub project libbrotli (https://github.com/bagder/libbrotli) a while ago to help build and install a _library_ for brotli encoding/decoding, using only code from this repository. libbrotli is only a meta-project with mostly autotools to build, install and package a "library" for brotli since this original brotli home does not provide that. It only uses compression/decompression source code from the brotli tree.

This concept seems to resonate with a decent amount of users who appreciate being able to get a library (or two actually) out of a build and install process for use in various projects.

I would prefer if this functionality was provided by the brotli project itself and I'll offer to merge/translate it over to a pull-request or something if you'll agree this is interesting. I'd prefer to remove myself as a middle man here.

Lukasa:
I'd like to add my voice to support this idea. As the author of the Python wrapper library [brotlipy](https://pypi.python.org/pypi/brotlipy), I have had to adopt the changes made by @bagder in order to extract the portions of the code that are relevant to Python library usage. This is problematic, because it forces me to make transformations on the code that make it incompatible with any brotli library installed from elsewhere, ensuring that I have to keep track of changes made in this repository rather than dynamically link against a binary library provided by the OS itself.

anthonyryan1:
I'd also like to express support for this idea.

The [brotli nginx module](https://github.com/google/ngx_brotli) currently depends on @badger's libbrotli because it can't use this.

If we want to see Brotli adoption pick up on the web, we need to get the web server plugins packaged on various Linux distros. I'm personally very eager to get nginx + Brotli ready on Gentoo, but I'd strongly prefer to have the library and the encoder tools in the same git tree and on the same release cycle.

fred-wang:
I'd also like to express support for a separate library. I've worked on WOFF2 support in Gecko and the WebKit GTK port and it would be more convenient to link to a system library instead of having duplicating code in the repos that we have to build & keep in sync.

jyrkialakuijala:
Please go ahead with transforming libbrotli to a pull request that merges libbrotli into brotli.

fred-wang:
With the new CMake build system it's now easy to create such brotli libraries. FYI I've opened #421.

fholzer:
@fred-wang but that doesn't mean that @bagder's  repo is no longer required, right? Built brotli with `-DBUILD_SHARED_LIBS=1` but nginx's `./configure` couldn't find the lib. Am i doing something wrong?

rockdaboot:
[...]

PR #332, opened 2016-03-20
build: create an encoder and decoder library

bagder:
Based on commit https://github.com/bagder/libbrotli/commit/e992cce7a17 after initial discussion in issue #326.

First take. This version unfortunately duplicate the list of sources from enc/Makefile and dec/Makefile, which I would rather avoid. It also doesn't use the `shared.mk` magic that the existing Makefiles use.

Run

```
$ ./autogen.sh
$ ./configue
$ make
$ make install
```

Any feedback on the approach? I decided to put the files in the root dir since that is a common convention. Let me know what you think.
# docs

`make install` puts the headers in $prefix/install/dec and $prefix/install/inc subdirectories to be used for applications.

This setup builds one encoder lib and one decoder lib. `libbrotlidec.pc` and `libbrotlienc.pc` are generated to help applications know how to link with them.

The `Makefile.am` file needs to be maintained and state the correct SONAME numbers for the generated shared libraries. They say 1.0.0. for both, but its really up to you guys to decide how to do them.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,0 +1,89 @@
+AUTOMAKE_OPTIONS = foreign nostdinc subdir-objects
+
+ACLOCAL_AMFLAGS = -I m4
+
+lib_LTLIBRARIES = libbrotlidec.la libbrotlienc.la
+
+DECODE = dec/bit_reader.c dec/decode.c dec/huffman.c dec/state.c	\
+  dec/dictionary.c
+
+ENCODE = enc/backward_references.cc enc/histogram.cc			\
+ enc/block_splitter.cc enc/literal_cost.cc enc/brotli_bit_stream.cc	\
+ enc/metablock.cc enc/encode.cc enc/static_dict.cc			\
+ enc/encode_parallel.cc enc/streams.cc enc/entropy_encode.cc		\
+ enc/dictionary.cc enc/utf8_util.cc enc/compress_fragment.cc		\
+ enc/compress_fragment_two_pass.cc
+
+DECODEHEADERS = dec/decode.h dec/state.h dec/types.h dec/bit_reader.h	\
+ dec/huffman.h dec/port.h dec/context.h dec/dictionary.h		\
+ dec/transform.h dec/prefix.h
+
+ENCODEHEADERS = enc/encode.h enc/command.h enc/hash.h enc/ringbuffer.h	\
+ enc/static_dict.h enc/streams.h enc/fast_log.h enc/prefix.h		\
+ enc/dictionary_hash.h enc/find_match_length.h enc/port.h		\
+ enc/transform.h enc/dictionary.h enc/backward_references.h		\
+ enc/histogram.h enc/block_splitter.h enc/metablock.h enc/context.h	\
+ enc/literal_cost.h enc/cluster.h enc/bit_cost.h enc/entropy_encode.h	\
+ enc/brotli_bit_stream.h enc/write_bits.h enc/static_dict_lut.h		\
+ enc/encode_parallel.h enc/types.h enc/utf8_util.h			\
+ enc/compress_fragment.h enc/compress_fragment_two_pass.h		\
+ enc/entropy_encode_static.h
+
+EXTRA_DIST = CONTRIBUTING.md README.md
+
+# install headers in $(includedir) with subdirs
+nobase_include_HEADERS = $(DECODEHEADERS) $(ENCODEHEADERS)
+
+pkgconfigdir = $(libdir)/pkgconfig
+pkgconfig_DATA = libbrotlienc.pc libbrotlidec.pc
+pkgincludedir= $(includedir)/brotli
+pkginclude_HEADERS = 
+
+LIBBROTLIENC_VERSION_INFO = -version-info 1:0:0
+LIBBROTLIDEC_VERSION_INFO = -version-info 1:0:0
+# This flag accepts an argument of the form current[:revision[:age]]. So,
+# passing -version-info 3:12:1 sets current to 3, revision to 12, and age to
+# 1.
+#
+# If either revision or age are omitted, they default to 0. Also note that age
+# must be less than or equal to the current interface number.
+#
+# Here are a set of rules to help you update your library version information: 
+#
+# 1.Start with version information of 0:0:0 for each libtool library.
+#
+# 2.Update the version information only immediately before a public release of
+# your software. More frequent updates are unnecessary, and only guarantee
+# that the current interface number gets larger faster.
+#
+# 3.If the library source code has changed at all since the last update, then
+# increment revision (c:r+1:a)
+#
+# 4.If any interfaces have been added, removed, or changed since the last
+# update, increment current, and set revision to 0. (c+1:r=0:a)
+#
+# 5.If any interfaces have been added since the last public release, then
+# increment age. (c:r:a+1)
+#
+# 6.If any interfaces have been removed since the last public release, then
+# set age to 0. (c:r:a=0)
+#
+
+libbrotlidec_la_CPPFLAGS_EXTRA = -DBROTLI_BUILDING_LIBRARY
+libbrotlidec_la_LDFLAGS = $(AM_LDFLAGS) $(LIBBROTLIDEC_VERSION_INFO) $(LDFLAGS)
+libbrotlidec_la_CFLAGS = $(AM_CFLAGS) $(libbrotli_la_CFLAGS_EXTRA)
+libbrotlidec_la_CXXFLAGS =
+libbrotlidec_la_CPPFLAGS = $(AM_CPPFLAGS) $(libbrotli_la_CPPFLAGS_EXTRA)
+libbrotlidec_la_SOURCES = $(DECODE) $(DECODEHEADERS)
+
+libbrotlienc_la_CPPFLAGS_EXTRA = -DBROTLI_BUILDING_LIBRARY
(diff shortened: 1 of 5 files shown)


nemequ:
Okay, now that #326 is closed I'm not sure how this should be handled.

There is a single pkg-config file (brotli.pc) for all three libraries (enc, dec, common) right now.  It wouldn't be hard to create a `brotli-common.pc` for `libbrotlicommon`, then create `brotli-{enc,dec}.pc`, and just have `brotli.pc` require the `brotli-{encoder,decoder}` packages.  We could also consider only using the encoder or decoder an "advanced" use case and simply not support if via pkg-config.  I don't really have a preference, but if @eustas wants to go with the 4 pkg-config files I don't mind putting together a quick patch (it would just require adding a few lines to `CMakeLists.txt`).

This PR also adds an autotools-based build system, which I consider a mistake.  Brotli already has CMake, premake, and bazel build systems, plus language-specific build systems for the Python and Java bindings.  Please, no more confusion.  To be fair, though, this PR may predate CMake support.

As for the supporting the standard `autogen.sh/configure && make && make install` flow (you separated `autogen.sh` and `configure`, but *usually* `autogen.sh` calls `configure`, so you have `autogen.sh` for git checkouts and `configure` for releases), I've previously mentioned the idea of using my [configure-cmake](https://github.com/nemequ/configure-cmake/) script to emulate this to avoid confusion (and provide a more comfortable way to configure the build; cmake's CLI is horrible).  I'm still willing to submit a PR for that if desired.

fred-wang:
@nemequ Regarding the pc, does have one or several files change how pkg-config will be used? If so, my guess would be to try and follow @bagder 's files so that people can keep using the same setup.

Regarding autotools and autogen, I would say these are not important features given that @eustas  decided to use different build systems.

nemequ:
> @nemequ Regarding the pc, does have one or several files change how pkg-config will be used?

(I'm not completely sure what you're asking here.  If I haven't answered below, could you re-phrase it a bit?)

Currently there is one pkg-config file (brotli.pc), which links to all three libraries (common, enc, dec).  So you can do

```sh
cc ... foo.c `pkg-config --libs --cflags brotli`
```

The most common use case will almost certainly be to link to all three libraries, so I think `brotli` should continue to act as it does.  If @eustas wants to include pkg-config support for linking to the libraries individually, I would suggest 4 files: `brotli.pc`, `brotli-common.pc`, `brotli-encoder.pc`, and `brotli-decoder.pc`.  The encoder and decoder packages would depend on common, and brotli would depend on both encoder and decoder (and common via those module's dependencies on it).  That would allow the existing behavior via the `brotli` module, or you could just use `brotli-encoder` or `brotli-decoder`:

```sh
cc ... unbrotli.c `pkg-config --libs --cflags brotli-decoder`
```

> If so, my guess would be to try and follow @bagder 's files so that people can keep using the same setup.

If you're saying the names should be. *e.g.*, `libbrotlidec.pc` instead of `brotli-decoder.pc`, most pkg-config files don't include the "lib" prefix (unless the project name includes it, such as "libpng") and try to provide "pretty" names instead of perfectly matching the shared library names.  Given that the names @badger suggested haven't been part of Brotli, much less an official release, at this point I'd much rather have something which matches pkg-config conventions than @badger's fork, especially considering how trivial it is to swap out the names.  From a technical perspective, though, the names don't really matter.

> Regarding autotools and autogen, I would say these are not important features given that @eustas decided to use different build systems.

A *lot* of people look for configure/autogen.sh scripts.  It's important enough that there is a configure script in the brotli repository which does nothing other than tell people what build systems are available.  I'm not suggesting a full autotools build system, just replacing that script with [configure-cmake](https://github.com/nemequ/configure-cmake/) to provide a more comfortable experience.  It's an easy way to "fix" that part of this issue and make brotli a bit more approachable.  Personally, it doesn't matter to me one way or the other; whatever @eustas wants to do.

Anyways, I just need a decision on what to do, then I can throw together a PR (or, of course, this issue can just be closed if you just want to keep everything as-is).

fred-wang:
> (I'm not completely sure what you're asking here. If I haven't answered below, could you re-phrase it a bit?)

I'm just asking if your patch allows to link to library individually but you seem to suggest that it does not below.

> The most common use case will almost certainly be to link to all three libraries, 

I don't know what is "most common" for you, but for example in web engines we don't care about the encoder library we just want the decoder to use the WOFF2 fonts.

> If you're saying the names should be. e.g., libbrotlidec.pc instead of brotli-decoder.pc, most 

No I was not talking about names, just about the features. If people are already linking library individually and your *.pc does not allow that, then I think it should be fixed.

> A lot of people look for configure/autogen.sh scripts.

I'm not saying autotools or autogen must not be used. I'm just saying that we already have some build systems and docs so that should be enough for people to figure out who to create encore/decoder libraries. Of course we can add more if people really want but that's a separate issue and not the goal of this pull request IIUC.

nemequ:
> I'm just asking if your patch allows to link to library individually but you seem to suggest that it does not below.

Ah.  Right, what is currently in Brotli doesn't provide a way to link to each library individually through pkg-config.  Obviously you can still link to the libraries directly, though.

> I don't know what is "most common" for you, but for example in web engines we don't care about the encoder library we just want the decoder to use the WOFF2 fonts.

I'm talking about projects, not user base.  Anecdotally (I'm *way* too lazy to try to do a real survey), I think a lot more projects will want to use both.

That said, I don't have any objection to the idea of adding a few more pkg-config to satisfy the unidirectional use case.  I just want to make sure that's what @eustas wants before I put together a PR so I don't waste effort.

> I'm not saying autotools or autogen must not be used. I'm just saying that we already have some build systems

You understand that I'm not suggesting adding an autotools build system, right?  configure-cmake is just [a shell script](https://github.com/nemequ/configure-cmake/blob/master/configure) you can drop into a CMake project (replacing the existing, somewhat useless, [configure](https://github.com/google/brotli/blob/master/configure) script) which maps an autotools-style interface to CMake; for example `../configure --prefix=/usr --disable-debug CC=clang` just calls something like `cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang`.  autogen.sh wouldn't be necessary for brotli.  In other words, the entire patch would replacing the `configure` script in brotli with the one in configure-cmake.

>  and docs so that should be enough for people to figure out who to create encore/decoder libraries.

CMake's CLI is a huge pain, and the documentation sucks.  `cmake --help` is almost completely useless, and it's impossible to be sure which variables effect the build without looking through `CMakeLists.txt`; non-brotli-specific docs may or may not apply and are incomplete and/or contain irrelevant information, and the docs in brotli's README are very incomplete (though better off than premake or bazel).

An autotools-style configure script, OTOH, has a useful `--help` and the options are standard and well-known to the point where many (most?) people won't even need documentation.

> Of course we can add more if people really want but that's a separate issue and not the goal of this pull request IIUC.

To be fair, I don't think pkg-config is really the point, either.  AFAICT the real point was creating and installing separate encoder/decoder libraries, which Brotli already does (now, not when this PR was filed).  Beyond that, the only stuff he mentioned was pkg-config and autotools.

Anyways, I don't really care about either issue.  IMHO adding some more pkg-config files and an autotools style interface for the build system are both good ideas, but neither effects me.  Whatever everyone else wants to do is fine.

fred-wang:
@nemequ : Apparently you are insisting a lot for stuff you said you do not care ;-)

Just to clarify my point: My understanding of the libbrotli project was indeed to allow brotli libraries to be created and used. So I see how the ability to generate shared and install libraries & pc files is important for that purpose and this is now possible (but not when this PR was opened). But I do not follow the connection between the librotli and the rest of what you are talking about, sorry...

nemequ:
> @nemequ : Apparently you are insisting a lot for stuff you said you do not care ;-)

That's why I mentioned that I don't really care; my point there was that I'm really not trying to be insistent about this.  Unfortunately, without body language it's hard for people to know that, while I do have opinions about how to proceed, I'm really not passionate about them.  These are fairly minor issues, and we're spending an inordinate amount of time talking about them.  I really just want to be done with it, especially considering I'm not really going to be using anything we're talking about here.  I'm really only here because I wrote the initial cmake (and pkg-config) stuff and I'm trying to help maintain it.

My positions are basically:

 * IMHO three more pkg-config files should be added, which I think we agree on.  However, I don't think it's a big deal if they aren't… pkg-config is nice to have but it's not critical, and I think the existing pkg-config file supports *most* users.
 * Adding configure-cmake would be nice because I think it would make building brotli easier for a lot of people, but if @eustas doesn't want do add it that's fine with me.
 * Those 3 pkg-config files should be called "brotli-{common,encoder,decoder}" not "libbrotli{common,enc,dec}" because that's the convention for pkg-config names, and I don't think it matters that those don't match @badger's original patch.  AFAICT, you disagree ("my guess would be to try and follow @bagder 's files so that people can keep using the same setup"), unless you've changed your mind.

fred-wang:
I'd recommend to close this pull request and to open separate issues / PRs for the different topics discussed here. In my opinion, only the discussion about pc files is relevant to the initial goal of having a brotli encoding/decoding library.

fred-wang:
@nemequ  In order to avoid hijacking this PR and to mix up too many topics, I opened #473 and #474 ; feel free to open more separate issues or to submit pull requests.

@eustas Can you please close this PR?

eustas:
Found out that trying to push CMake towards GNU builds is too painful.
Could you take a look at #620, please

PR #334, opened 2016-03-24
Fix build with -Wshadow.

PiotrSikora:
Local variable shadowing makes code quite confusing in most cases.

I'm not sure if the new variable names are correct in all those cases,
so feel free to rename them.


eustas:
Fixed in v0.5

PR #336, opened 2016-04-10
appveyor: also upload source distribution packages to Github Releases

anthrotype:
Currently, only the binary wheel packages are uploaded to the Github Releases page on a new tag.
This patch allows to upload the source distributions (*.tar.gz and *.zip) as well.

The Github-generated zip of the repository can already be pip-installed as it is. However, in order for the pip `--find-links` mechanism to work properly, a zip has to be named in a way that pip recognizes as a Python source package.
I believe it's better to have setuptools create the package (via `sdist` command) and then upload it alongside the binary wheels.

Once the source packages are uploaded to Github with the next tag, users will be able to use a single command to install brotli on any platform, whether or not pre-compiled *.whl packages are available that match the user's python version and architecture:

    pip install --find-links https://github.com/google/brotli/releases/latest brotli

Pip will automatically choose the correct binary package, or if that isn't available it will use the source distribution, instead of failing as it happens now, for example, when running the above command on a Linux system, for which binary wheels for brotli are not available yet ([binary wheels for Linux](https://github.com/pypa/manylinux) are possible but it's tricky. I could give a try later).

For example, say I'm testing my python package that requires brotli on Travis on both Linux and OS X. When testing on OS X, I would like to fetch the pre-compiled wheel to speed up the build, while on Linux I can have it built from source. If I use the above command, pip will do the rest. Currently, I would have to pass different pip commands for each environment, or I would have to always pip install from a cloned git repository, even when a wheel is available.

Even better, of course, if we uploaded brotli to the Python Package Index (#72).


anthrotype:
ping @eustas

anthrotype:
Thanks!

PR #380, opened 2016-06-17
Fix appveyor build

eustas:


anthrotype:
the problem is `pip wheel` command is trying to create a default "build" directory in the root of the repo, but there's a file called BUILD. You can try to pass the `-b`, or `--build <dir>` option to `pip wheel` command to use a different directory where to build into.

eustas:
Tried to add "-b bin" to "pip wheel", it still fails.

anthrotype:
oh I see that.. hm that's weird. Try adding a `-v` verbose flag maybe.

Or also you can try calling the `bdist_wheel` command directly instead of through pip:
```
python setup.py bdist_wheel -b bin -d dist
```

anthrotype:
I was able to replicate the issue locally on my VM.
It looks like a bug in either `setuptools` or `wheel` or `pip` (or all of them!).
With this command line I was able to make it work:
```
python setup.py build -b bin bdist_wheel -d dist
```

It looks like the bdist_wheel command in turns calls the build command which somehow always defaults to create a `build` directory, despite we pass `-b bin` to bdist_wheel...

anthrotype:
yeah, the setup.cfg trick should also fix it :+1:

eustas:
Going to prepare a separate PR, to remove intermediate commits

PR #381, opened 2016-06-17
Fix wheels build

eustas:


eustas:
Travis "after_success" feels better, still appveryor is unhappy.

eustas:
Finally, appveryor is green

PR #387, opened 2016-06-27
Add CMake buildfiles. Use CMake in travis.

eustas:


CMakeLists.txt
```diff
@@ -0,0 +1,140 @@
...
+
+add_library(brotli_common STATIC
+  common/dictionary.c)
+add_library(brotli_dec STATIC
+  dec/bit_reader.c
+  dec/decode.c
+  dec/huffman.c
+  dec/state.c)
+add_library(brotli_enc STATIC
+  enc/backward_references.c
```
eustas: Why do not use GLOB? (And have policy that common/, dec/ and enc/ may contain only files to be built)

eustas:
Travis looks happy, indeed.
The only thing left to do is to sign CLA (sorry for the bureaucracy).

CMakeLists.txt
nemequ: > Why do not use GLOB? (And have policy that common/, dec/ and enc/ may contain only files to be built)

From the CMake documentation:

> We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.

I can change it to use glob if you want, but with CI in place I don't think there is much danger of forgetting to add/remove a file, so IMHO it is better to leave as-is.

nemequ:
:o didn't realize you opened a PR for this.  I signed the CLA for a patch to Gipfeli, so that shouldn't be an issue anymore.

Pretty much everything should be working now, the only real exception is TSan, which I don't think is going to happen on Ubuntu 12.04.

The Travis error above looks transient (it happens sometimes); asking Travis to rebuild the commit should work.

eustas:
Don't worry about CLA. Rerunning travis...
Thank you for this PR!

eustas:
It looks like problem with gcc4.5 persists...

nemequ:
That's very odd.  The same commit works for me: https://travis-ci.org/nemequ/brotli/jobs/146810092

nemequ:
Disabled gcc-4.5, everything finally seems to be working.  The failure with ubsan is expected; there is some undefined behavior in brotli (even with `-DBROTLI_BUILD_PORTABLE`).

Still not sure why gcc-4.5 works in my repo but not as a PR, but I guess it's not a big deal.

eustas:
Going to calm down clang-ubsan in few moments.

eustas:
Ubsan is happy now.

eustas:
I signed it!

nemequ:
I signed it!

nemequ:
I'm guessing if **I** opened up a PR it would resolve the CLA thing…  would that help?

eustas:
I think - yes. This bot is always harassing me...

nemequ:
Argh, doesn't look like GitHub will let me file a duplicate PR; it just links me to this one.  I guess if you close this one I could probably do it…

eustas:
Okay, let's try

PR #388, opened 2016-06-28
fix build on latest msys2 / mingw64 x86_64

TYoung86:


TYoung86:
Spamming buttons before I was ready, appologies.

PR #390, opened 2016-07-06
Add a "lib" target to the Makefile to build a static library: libbrotli.a

chad-iris:
This small change to the Makefile supports building a libbrotli.a archive of the encoding and decoding objects (not including the CLI tool object or running the test suite).  This provides an easy target for linking when brotli code is included in projects that use the encoding and decoding components.  This is only an extension, the previous behavior for other targets remain.

P.S. I like the approach of providing a simple, portable Makefile for building when an autoconfig/configure/make can be avoided.  This simplicity makes incorporation of brotli into small projects (that do not have their own autoconf/etc.) that much easier.  Thanks!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,11 @@
 OS := $(shell uname)
-SOURCES = $(wildcard common/*.c) $(wildcard dec/*.c) $(wildcard enc/*.c) \
-          tools/bro.c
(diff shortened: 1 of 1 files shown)


chad-iris:
I signed it!

eustas:
Hello. Sorry for the super-late response.

 Thank you for your PR.

PR #391, opened 2016-07-12
Move public headers to the "include" directory and expose them.

PiotrSikora:
Public headers are now available via:                                         

        #include <brotli/decode.h>
        #include <brotli/encode.h>
        #include <brotli/types.h>
                                                                                      
        #include <brotli/compressor.h>
        #include <brotli/streams.h>

The exposure of "types.h" (which is only useful for MSVC builds) is
a bit unfortunate, but this change would be much bigger without it.

Makefile tested with "make test", BUILD file tested with ngx_brotli,
setup.py tested with "python setup.py build_ext test", premake5.lua
is broken, but it was broken before this change.


PiotrSikora:
Please note that C++ interface wasn't very well tested, since there are no consumers of it.

eustas:
Hello, Piotr.
  Please, give me few days to public new decoder API first...

PiotrSikora:
Hey Evgenii,
should I rebase this on top of new master? Are the changes proposed in this pull request acceptable?

PR #399, opened 2016-07-29
Add mingw support.

nemequ:
See https://travis-ci.org/quixdb/squash/jobs/148073219 for the failures with the current code.  Figured I may as well add a Travis build (including running the tests in wine) while I was at it.


nemequ:
Pushed a second commit which should, I *think*, resolve the issues where Travis works for nemequ/brotli but not google/brotli.  It should also speed up the build a bit.

Issue #374, opened 2016-06-14
Provide a build build system for both system-wide installation and bundling


nemequ:
Since [Squash](https://quixdb.github.io/squash/) bundles Brotli in our tree, we basically maintain a separate build system for Brotli.  I'm trying to reduce the amount of duplication between Squash and the libraries it bundles, so I'd like to be able to reuse Brotli's build system, which would mean making it compatible with our own (CMake-based) build-system.  Obviously such a system wouldn't be specific to Squash; any project which uses CMake and wants to bundle Brotli could benefit from it.  With such a system, bundling Brotli would be as simple as something like:

```cmake
set(BROTLI_BUNDLE_MODE ON)
add_subdirectory(brotli)

target_link_libraries(mycode brotli)
```

Brotli would then be built as part of the parent project, including unit tests (which could also automatically integrate with ctest, so running `make test` would run brotli's tests as well) and documentation (if applicable).

The rest of this is largely copied from quixdb/squash#202, edited slightly to be specific to brotli.

From Squash's perspective, as well as anyone else who wants to bundle the library, what we need is a way to build a static library (or two, if you want to separate the encoder and decoder), disable any installation, and control whether or not to have CTest run unit tests.  Finally, it would be great if we could simply reference a target and have it automatically set any necessary flags (CMake ≥ 2.8.11 can do this, you just need to set the proper `INTERFACE_*` target properties).

If you would like to see an example of a system which works the way we want, take a look at [BriefLZ's](https://github.com/jibsen/brieflz/blob/master/CMakeLists.txt).  I'm willing to create a similar system for Brotli, though obviously it would need to be a bit more extensive since Brotli is a larger project.  I don't want to fork anything, so if you aren't interested we'll just stick with the current system for that plugin.  Obviously not everyone is familiar with CMake so part of this is that I would help with maintenance as needed.  However, before I proceed I need a few questions answered:

First, are you willing to provide a CMake-based build system in your tree?  If so, are you okay with making it flexible enough to meet our needs as described above?  If no to either of those, you can go ahead and close this issue and we'll keep interacting with Brotli just as we do today.  Otherwise, there are a few ways to put this together:

1. Place a single CMakeLists.txt file in the top level of your project, which contains all the logic for your entire project.  If you're familiar with autotools, this is like a non-recursive build system.
[...]

PR #400, opened 2016-07-29
appveyor: add Visual Studio builds (based on CMake)

nemequ:
This only goes back to VS 12 (2013) because MSVC didn't support log2 until then.

I think I'm done now, unless you want me to do some of the stuff from #374 (pkg-config, install a library, configure-cmake, etc.).


nemequ:
I added another commit to this which is unrelated to the PR title.  Apparently some versions of CMake will cache the result of the check for log2, so if it fails without linking to libm it would simply skip the test *with* libm.  Changing the name of the variable for the second test (with libm) works around the issue.

The issue with Python on AppVeyor seems to have been caused by a change in AppVeyor; builds which previously worked don't anymore.  I'll submit a bug report with them and try to figure out what is going on there…

anthrotype:
It looks like it's a bug in the latest setuptools not correctly enquoting library paths with spaces.
Maybe you can temporarily try to download an earlier version of setuptools which was known to be workong.
https://github.com/pypa/setuptools/issues/694

nemequ:
Yes, https://github.com/FeodorFitsner/brotli/blob/master/appveyor.yml#L31 fixes the issue.  I'd just like to figure out if we should bother or just wait for setuptools to tag a new release…

anthrotype:
Let's wait, I guess. They are usually pretty quick.

anthrotype:
hmm.. looks like it's not fixed yet:
https://github.com/pypa/setuptools/issues/694#issuecomment-237873988

I think it's a good idea to pin down a specific version of setuptools, instead of always upgrading to the latest at every appveyor build. We could only update setuptools when needed and when we know does not break anything else.

anthrotype:
The latest setuptools v25.1.6 should fix the issue.
https://pypi.python.org/pypi/setuptools/25.1.6

@nemequ can you try to re-run the Appveyor build?

nemequ:
I can't rebuild the PR without changing the commits; @eustas would need to do that.  However, it works in my repository now (see https://ci.appveyor.com/project/quixdb/brotli/build/1.0.26).

eustas:
Appveyor seems to be happy now. Will merge soon.

PR #401, opened 2016-07-30
Fix link in Readme

mehdi-farsi:
As he said:

```
That's one small step for a man, one giant leap for mankind.
```

PR #402, opened 2016-07-31
draft-alakuijala-brotli is now RFC 7932

nextPrime:
So change the link to https://tools.ietf.org/html/rfc7932


nextPrime:
Would conflict with #401.

PR #406, opened 2016-08-10
Fix 3.x python deployment

eustas:


anthrotype:
you shouldn't use `pip3`, but activate the relevant python environment before calling the respective `pip` command.

eustas:
Please, could you help with those things?

anthrotype:
Sure, I'll do it later today.

PR #407, opened 2016-08-10
Fix python build

anthrotype:
We need to source the virtual environment at the `script` and `after_success` stages inside`.travis.sh` script.

We can build and deploy only OSX wheels for now, but I will work on adding `manylinux1` support later.

There's no need to build wheels using homebrew python (64bit only), as the wheels produced by Mac Python (from Python.org) are "universal", so they will work on homebrew python as well.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 case "$1" in
     "install")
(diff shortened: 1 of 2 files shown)


anthrotype:
should be good to merge now.

nemequ:
Sorry, I guess I screwed this up when I added the CMake builds :(

Regarding "source $file", the standard version is ". $file".  Doesn't really matter if .travis.sh since that's only supposed to run on Travis, though.

PR #416, opened 2016-08-22
Remove some deadcode

sylvestre:


eustas:
Nice catch 👍 
It is dead since last October, but all compilers remain silent...
TravisCL experience some troubles, but I'll rerun the build when it feel better.

sylvestre:
seems it feel better :)

PR #418, opened 2016-08-23
Bazel: use cc_library instead of cc_inc_library.

PiotrSikora:
cc_inc_library is broken when used with external repositories
(see: https://github.com/bazelbuild/bazel/issues/1596), which
makes it a bit useless at the moment.

Switch to using cc_library with "includes" attribute to expose
public headers.

While there, fix order of attributes in ":brotli_common" target.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -56,18 +56,19 @@ filegroup(
     srcs = glob(["enc/*.c"]),
 )
 
-cc_inc_library(
+cc_library(
     name = "brotli",
     hdrs = [":public_headers"],
-    prefix = "include",
+    copts = STRICT_C_OPTIONS,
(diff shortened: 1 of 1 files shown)


eustas:
Please consider adding TODO / creating issue to use cc_inc_library when it is fixed.
Using "includes" causes warning:
`WARNING: /usr/local/google/home/eustas/clients/brotli/BUILD:63:16: in includes attribute of cc_library rule //:brotli: 'include' resolves to 'include' not in 'third_party'. This will be an error in the future.`

PiotrSikora:
Unfortunately, this warning is also known issue: https://github.com/bazelbuild/bazel/issues/1286

PR #419, opened 2016-08-23
Bazel: temporarily disable ":bro" target.

PiotrSikora:
Bazel doesn't like "tools" directory, which makes the whole
BUILD file unusable.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -86,13 +86,13 @@ cc_library(
     deps = [":brotli_common"],
 )
 
-cc_binary(
-    name = "bro",
-    srcs = ["tools/bro.c"],
-    copts = STRICT_C_OPTIONS,
-    linkstatic = 1,
-    deps = [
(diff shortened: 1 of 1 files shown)


eustas:
Works for me both with 0.3.0 and 0.3.1 versions.
Could you publish verbose Bazel report. Perhaps there is some other bug in Bazel, or some environment / configuration problem.

PiotrSikora:
After wiping everything Bazel-related from my systems (since it was broken on all of them), I found the issue.

It was outdated `~/.bazelrc` file generated by older version of Bazel:

    build --package_path %workspace%:/path/to/bazel/base_workspace
    fetch --package_path %workspace%:/path/to/bazel/base_workspace
    query --package_path %workspace%:/path/to/bazel/base_workspace

Thanks for double-checking and sorry for the noise!

PR #421, opened 2016-08-27
Add support for CMake's BUILD_SHARED_LIBS option. #326

fred-wang:
This changes allows to build a small brotli binary that relies on shared libraries dec, enc and common (instead of a single binary). In particular this will allows distro maintainers to create independent libbrotli libraries (issue #326).


fred-wang:
I signed it!

eustas:
Nice! Thanks

PR #424, opened 2016-08-30
check for __ARM64_ARCH_8__ in dec/port.h so that arm64 arch under cla…

mdejong:
…ng is detected, check for __ARM_ARCH being exactly equal to 7 so that arm64 arch does not define BROTLI_TARGET_ARMV7

### DIFF
--- a/dec/port.h
+++ b/dec/port.h
@@ -32,13 +32,13 @@
 #include "../common/port.h"
 
 #if defined(__arm__) || defined(__thumb__) || \
-    defined(_M_ARM) || defined(_M_ARMT)
+    defined(_M_ARM) || defined(_M_ARMT) || defined(__ARM64_ARCH_8__)
 #define BROTLI_TARGET_ARM
-#if (defined(__ARM_ARCH) && (__ARM_ARCH >= 7)) || \
-    (defined(M_ARM) && (M_ARM >= 7))
+#if (defined(__ARM_ARCH) && (__ARM_ARCH == 7)) || \
+    (defined(M_ARM) && (M_ARM == 7))
 #define BROTLI_TARGET_ARMV7
 #endif  /* ARMv7 */
-#if defined(__aarch64__)
+#if defined(__aarch64__) || defined(__ARM64_ARCH_8__)
 #define BROTLI_TARGET_ARMV8
 #endif  /* ARMv8 */
 #endif  /* ARM */


mdejong:
I signed it

My github username is mdejong.

Mo DeJong

eustas:
The only difference between v7 and v8 in decoder is using "rbit".
Could you check that "rbit" works on ARMv8, please.

eustas:
(and 32 vs 64 bit, of course)

mdejong:
So, I can confirm that rbit would compile on arm64, but it causes a runtime error. It does compile, the generated asm on an arm64 target looks like:

```
0x1000a8f7c <+1208>: str    w9, [sp, #100]
0x1000a8f80 <+1212>: ldr    w9, [sp, #100]
0x1000a8f84 <+1216>: rbit   x9, x9
0x1000a8f88 <+1220>: str    w9, [sp, #96]

```

For example, you could make this compile after adding my patch by making this change:

```
#if BROTLI_MODERN_COMPILER || defined(__llvm__)
#if defined(BROTLI_TARGET_ARMV7) || defined(BROTLI_TARGET_ARMV8)
static BROTLI_INLINE unsigned BrotliRBit(unsigned input) {
  unsigned output;
  __asm__("rbit %0, %1\n" : "=r"(output) : "r"(input));
  return output;
}
#define BROTLI_RBIT(x) BrotliRBit(x)
#endif  /* armv7 */
#endif  /* gcc || clang */

```

The change I made was to just get BROTLI_TARGET_ARMV7 to only be defined for the ARM 7 targets while keeping it from being defined for arm64 which is v8.

mdejong:
After a bit of poking around, I found the following: https://patchwork.kernel.org/patch/5144161/

Changing BrotliRBit() to the following fixed the asm call to rbit on arm64:

```
#if BROTLI_MODERN_COMPILER || defined(__llvm__)
#if defined(BROTLI_TARGET_ARMV7) || defined(BROTLI_TARGET_ARMV8)
static BROTLI_INLINE unsigned BrotliRBit(unsigned input) {
  unsigned output;
#if defined(BROTLI_TARGET_ARMV8)
  __asm__("rbit %w0, %w1" : "=r" (output) : "r" (input));
#elif defined(BROTLI_TARGET_ARMV7)
  __asm__("rbit %0, %1\n" : "=r"(output) : "r"(input));
#endif
  return output;
}
#define BROTLI_RBIT(x) BrotliRBit(x)
#endif  /* armv7 || arm64 */
#endif  /* gcc || clang */

```

With this change, both arm 32 and 64 bit targets compile and execute correctly.

mdejong:
Hey guys, can this first commit be added to the repo? I can submit a second pull req for the second change related to the rbit instruction, but it depends on the first one.

eustas:
Hello. Hasn't #431 fixed the rbit problem?

mdejong:
The rbit issue is an additional change that can be added on after this pull req. I proposed a rbit modification also (see above), but this pull req stands on its own.

eustas:
Hello @mdejong. Please, could you check that replacing
`#if defined(BROTLI_TARGET_ARMV7)` with
`#if defined(BROTLI_TARGET_ARMV7) || defined(BROTLI_TARGET_ARMV8)`
before `BrotliRBit` declaration do not break build / tests?

mdejong:
I just created 459 and verified that rbit is executing properly when built on ARM64 with that change.

PR #426, opened 2016-09-14
Create large window brotli, new incompatible compression format.

Zip753:
Large window brotli is a brotli version that can use 30-bit (1GB) window. It uses the reserved short-code `0010001` to indicate 30-bit window, therefore breaking [section 9.1 of RFC 7932](https://tools.ietf.org/html/rfc7932#section-9.1).

This is an experimental change, so that brotli can be more competitive on large inputs.

The compression ratio on enwik9 for quality 11 went from 4.47 for 24-bit window to 5.02 for 30-bit window. Also, on [10GB corpus](http://mattmahoney.net/dc/10gb.html) the compression ratio for quality 10 and 11 increased from 2.52 and 2.54 to 2.78 and 2.80 respectively.

Additionally, large window brotli has been very useful in researching brotli distance encoding and exploring patching possibilities.

### DIFF
--- a/common/constants.h
+++ b/common/constants.h
@@ -32,10 +32,12 @@
 #define BROTLI_NUM_DISTANCE_SHORT_CODES 16
 #define BROTLI_MAX_NPOSTFIX 3
 #define BROTLI_MAX_NDIRECT 120
-/* BROTLI_NUM_DISTANCE_SYMBOLS == 520 */
+#define BROTLI_MAX_DISTANCE_BITS 30U
+/* BROTLI_NUM_DISTANCE_SYMBOLS == 616 */
 #define BROTLI_NUM_DISTANCE_SYMBOLS (BROTLI_NUM_DISTANCE_SHORT_CODES + \
                                      BROTLI_MAX_NDIRECT +              \
-                                     (24 << (BROTLI_MAX_NPOSTFIX + 1)))
+                                     (BROTLI_MAX_DISTANCE_BITS <<      \
+                                      (BROTLI_MAX_NPOSTFIX + 1)))
 
 /* 7.1. Context modes and context ID lookup for literals */
 /* "context IDs for literals are in the range of 0..63" */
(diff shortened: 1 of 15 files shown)


Zip753:
I signed it!

dsnet:
Hi, I'm just curious what the end goal of the large-window Brotli. If the experiment shows that large-window Brotli is competitive, will this functionality be merged into `master`, giving users the ability to choose a "large-window" setting?

I'm okay with seeing a new RFC be published that extends the current Brotli spec, but I'm concerned if the canonical Brotli implementation provides settings where people can output an invalid version of Brotli. I'm concerned about bad user experience where Brotli that this repo produced can't be decompressed by a RFC-compliant decoder implementation.

P.S. I'm the author of the [Go implementation of Brotli decoder](https://github.com/dsnet/compress/tree/master/brotli).

Zip753:
Hi, the goal of the large window brotli is mostly research-oriented, it's mostly used for facilitating distance encoding research. There are couple of other research tools that use large window brotli, I'm planning on publishing them soon as well.

Just to make it clear, this functionality isn't going to be merged into `master` anytime soon, especially since the new RFC was recently published. Large window brotli is one way compatible with RFC 7932, i.e. it can decompress classic brotli files, but not vice versa. So the main thing to acknowledge is that the output of this branch's encoder isn't brotli, therefore no one should try to decompress it with classic decoder.

jyrkialakuijala:
Please do not try to reproduce this experiment in other languages, we will keep making changes to the large window brotli experiment at will for the next 18 months or so -- also, do not use the large window brotli for actually storing data for now. While the current version gives ~10 % compression density improvement for enwik9 and even more in some special compression scenarios like patching, we have many further ideas on how to improve on this. This means that the format is currently not finalized, and will be changed many times (and always in an backward/forward incompatible way) before it will be even considered for integration into the brotli master repository -- or split into a separate repository.

dsnet:
> Please do not try to reproduce this experiment in other languages, we will keep making changes to the large window brotli experiment.

Don't worry, I wont. I'll implement it only if the work gets drafted as an RFC spec. I am excited to see where these experiments go. Good work.

PR #427, opened 2016-09-15
Add backward reference research tools to brotli repository.

Zip753:
This is a small set of simple tools used together with [large-window brotli](https://github.com/google/brotli/pull/426) to visualize backward reference distance distribution. You can find a description [here](https://github.com/Zip753/brotli/blob/master/research/README.md).


eustas:
E.g. link research/third_party/esaxx to https://github.com/hillbig/esaxx

PR #437, opened 2016-10-07
Also install the brotli headers when building the shared libraries. #326

fred-wang:


fred-wang:
@eustas Can you please take a look to that one-line change? I think this is the remaining important bit to fix issue #326 . The travis failures seem to happen on BUILD_SYSTEM=python, so that should be unrelated to CMake changes?

eustas:
I think it should be:
```
  if(NOT WIN32)
    install(DIRECTORY ${BROTLI_INCLUDE_DIRS}/brotli DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
  endif()
```

fred-wang:
@eustas OK, that works for me too (let's see later if some Cygwin or MSYS2 maintainers complain...). I updated the pull request.

eustas:
I meant, it should be outside of "is shared libs", because when linking statically, includes required as well =)

fred-wang:
@eustas OK, but at the moment we do not install the libraries at all when building in static mode, so the headers are not going to be used anyway. Maybe we want to remove the whole if(BUILD_SHARED_LIBS) here so that we also install the lib & include in static mode too?

eustas:
Sounds logical.

fred-wang:
@eustas OK, done.

eustas:
One last thing (sorry): any reason not to use paths provided by GNUInstallDirs.
I'm not very experienced in that matter, but it feels like this will make build a bit more portable.

eustas:
And CMake wants `ARCHIVE DESTINATION` as well...

fred-wang:
@eustas updated the install directory, but not sure what you mean with your last comment?
(PS: feel free to take over this work if that's faster, I'm not experienced with all these cmake build configs either...)

PR #445, opened 2016-10-17
Make BUILD_SHARED_LIBS default to ON. #326

fred-wang:
@eustas Let's see if that makes the bots happy (maybe we want to change the travis / appveyor files too...)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,8 @@ cmake_minimum_required(VERSION 2.8.6)
 
 project(brotli)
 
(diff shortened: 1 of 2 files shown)


eustas:
VS thinks that brotlienc depends on static brotlicommon...
Of course we can modify appveyor config to force building static artefacts, but first we should teach CMake to generate valid shared library build for VS.

eustas:
Working on DLL imports/exports + stub lib generation. That will include this PR.

fred-wang:
Closing as it is included in #450

Issue #191, opened 2015-10-01
Python binding: add streaming compression / decompression support.


rpasta42:
I'm trying to implement streaming compression using brotli. There's very little documentation and I haven't found anyone else trying to accomplish same thing. I assume I need to specify lgwin to set the block size, but when I run this code it gives me an error "brotli.error: BrotliDecompress failed"

my_str = 'very long string blah blah blah'
a = brotli.compress(my_str, lgwin=16)
b = brotli.decompress(a[0:16])

Am I doing something wrong or did I misunderstand the purpose of the lgwin parameter?

eustas:
Currently python brotli module doesn't support streaming compression/decompression.
lgwin is a parameter that is used to specify the length of LZ77 sliding window size.

Of course, streaming support will be eventually added to the Python bindings,... So stay tuned.

Kijewski:
> Of course, streaming support will be eventually added to the Python bindings,... So stay tuned.

Do you have a more precise ETA by now? More precise than "eventually" I mean.

KostyaKow:
I know C++ and Python, but have no familiarity with brotli codebase.

Where would I begin if I wanted to implement the Python binding for this?

thedrow:
Looks like this is complete. Should we close this issue?

PR #448, opened 2016-10-18
Python: Support streamed compression with the Compressor object

nicksay:
Add `flush` and `finish` methods to the `Compressor` object in
the extension module and change `compress` to only process data.
Now, one or more `compress` calls followed by a `finish` call
will be equivalent to a module-level `compress` call.

Note: To maximize the compression efficiency (and match
underlying Brotli behavior, the `Compressor` object `compress`
method does not guarantee all input is immediately written to
output. To ensure immediate output, call `flush` to manually
flush the compression buffer. Extraneous flushing can increase
the size, but may be required when processing streaming data.

---

_edited Oct 19: see updated description in commit and the thread below_

### DIFF
--- a/python/_brotli.cc
+++ b/python/_brotli.cc
@@ -2,7 +2,6 @@
 #include <Python.h>
 #include <bytesobject.h>
 #include <structmember.h>
-#include <cstdio>
 #include <vector>
 #include "../common/version.h"
 #include <brotli/decode.h>
@@ -86,6 +85,38 @@ static int lgblock_convertor(PyObject *o, int *lgblock) {
   }
 
   return 1;
+}
+
+static BROTLI_BOOL compress_stream(BrotliEncoderState* enc, BrotliEncoderOperation op,
+                                   std::vector<uint8_t>* output, uint8_t* input, size_t input_length) {
+  BROTLI_BOOL ok = BROTLI_TRUE;
+
+  size_t available_in = input_length;
+  const uint8_t* next_in = input;
+  size_t available_out = 0;
+  uint8_t* next_out = NULL;
+
+  while (ok) {
+    ok = BrotliEncoderCompressStream(enc, op,
+                                     &available_in, &next_in,
+                                     &available_out, &next_out, NULL);
+    if (!ok)
+      break;
+
+    size_t buffer_length = 0; // Request all available output.
+    const uint8_t* buffer = BrotliEncoderTakeOutput(enc, &buffer_length);
+    if (buffer_length) {
+      (*output).insert((*output).end(), buffer, buffer + buffer_length);
+    }
+
+    if (available_in || BrotliEncoderHasMoreOutput(enc)) {
+      continue;
+    }
+
+    break;
+  }
+
+  return ok;
 }
 
 PyDoc_STRVAR(brotli_Compressor_doc,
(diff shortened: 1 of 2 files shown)


nicksay:
@eustas 

Please take a look at the `compress_stream` function in particular to make sure it's sane.  The `Compressor` object now has 3 methods, each of which call `compress_stream` and the underlying `BrotliEncoderCompressStream` with the following operations:
- `compress` <=> `BROTLI_OPERATION_PROCESS`
- `flush` <=> `BROTLI_OPERATION_FLUSH`
- `finish` <=> `BROTLI_OPERATION_FINISH`

Looking at that list now, perhaps we should rename `Compressor.compress` to `Compressor.process`.  If we did that, the Python module would look like:

```py
# pseudo-code

class Compressor:
  """Class to do streamed compression across multiple calls, etc."""
  def process(s):
    # do BrotliEncoderCompressStream(BROTLI_OPERATION_PROCESS, s)
  def flush
    # do BrotliEncoderCompressStream(BROTLI_OPERATION_FLUSH)
  def finish
    # do BrotliEncoderCompressStream(BROTLI_OPERATION_FINISH)

def compress(s):
  """Function to do single-call compression, as exists now"""
  compressor = Compressor()
  return compressor.process(s) + compressor.finish()
```

Thoughts?

python/_brotli.cc
```diff
@@ -88,6 +87,50 @@ static int lgblock_convertor(PyObject *o, int *lgblock) {
...
+
+  size_t buffer_length;
+  if (input_length > 0) {
+    buffer_length = input_length + (input_length >> 2) + kBufferPadding;
+  } else {
+    input_length = 0;
+    input = NULL;
+    buffer_length = kBufferSize;
+  }
+  uint8_t* buffer = new uint8_t[buffer_length];
```
eustas: I think using `BrotliEncoderTakeOutput` is preferable in wrappers workflow.
It has 2 useful features:
* choosing size, allocating and freeing intermediate output buffer is not required anymore
* one less memcpy (from internal brotli buffer to intermediate output buffer)

nicksay: I pushed a new temporary commit using `BrotliEncoderTakeOutput`, but I'm getting errors when trying to execute sequences of operations.

Here are some scenarios:
```
process("one two three ")
finish()
```
**success**

```
process("one ")
process("two ")
process("three ")
finish()
```
**success**

```
process("one ")
process("two ")
flush()
process("three ")
```
**failure:** `BrotliEncoderCompressStream` returns `false`

```
process("one ")
process("two ")
process("three ")
flush()
finish()
```
**failure:** `BrotliEncoderIsFinished` returns `false` and output is incomplete

So it seems to me that either
1. something with `BrotliEncoderCompressStream(..., BROTLI_OPERATION_FLUSH, ...)` and `BrotliEncoderTakeOutput` is not working correctly
2. I'm calling `BrotliEncoderCompressStream` incorrectly now.

nicksay: I'm going to copy this comment to the new commit.

python/_brotli.cc
```diff
@@ -89,45 +89,32 @@ static int lgblock_convertor(PyObject *o, int *lgblock) {
...
     size_t available_in = input_length;
     const uint8_t* next_in = input;
-    size_t available_out = buffer_length;
-    uint8_t* next_out = buffer;
+    size_t available_out = 0;
 
     ok = BrotliEncoderCompressStream(enc, op,
                                      &available_in, &next_in,
-                                     &available_out, &next_out, 0);
+                                     &available_out, NULL, NULL);
```
nicksay: _comment copied from previous commit_

I pushed a new temporary commit using `BrotliEncoderTakeOutput`, but I'm getting errors when trying to execute sequences of operations.

Here are some scenarios:
```
process("one two three ")
finish()
```
**success**

```
process("one ")
process("two ")
process("three ")
finish()
```
**success**

```
process("one ")
process("two ")
flush()
process("three ")
```
**failure:** `BrotliEncoderCompressStream` returns `false`

```
process("one ")
process("two ")
process("three ")
flush()
finish()
```
**failure:** `BrotliEncoderIsFinished` returns `false` and output is incomplete

So it seems to me that either
1. something with `BrotliEncoderCompressStream(..., BROTLI_OPERATION_FLUSH, ...)` and `BrotliEncoderTakeOutput` is not working correctly
2. I'm calling `BrotliEncoderCompressStream` incorrectly now.

python/_brotli.cc
```diff
@@ -88,6 +87,37 @@ static int lgblock_convertor(PyObject *o, int *lgblock) {
...
+  BROTLI_BOOL ok = BROTLI_TRUE;
+
+  while (ok) {
+    size_t available_in = input_length;
+    const uint8_t* next_in = input;
+    size_t available_out = 0;
+
+    ok = BrotliEncoderCompressStream(enc, op,
+                                     &available_in, &next_in,
+                                     &available_out, NULL, NULL);
```
nicksay: _comment copied from previous commit_

I pushed a new temporary commit using `BrotliEncoderTakeOutput`, but I'm getting errors when trying to execute sequences of operations.

Here are some scenarios:
```
process("one two three ")
finish()
```
**success**

```
process("one ")
process("two ")
process("three ")
finish()
```
**success**

```
process("one ")
process("two ")
flush()
process("three ")
```
**failure:** `BrotliEncoderCompressStream` returns `false`

```
process("one ")
process("two ")
process("three ")
flush()
finish()
```
**failure:** `BrotliEncoderIsFinished` returns `false` and output is incomplete

So it seems to me that either
1. something with `BrotliEncoderCompressStream(..., BROTLI_OPERATION_FLUSH, ...)` and `BrotliEncoderTakeOutput` is not working correctly
2. I'm calling `BrotliEncoderCompressStream` incorrectly now.

eustas: Going to investigate that.

eustas: Found the problem: "take output" does not transition out from "flush requested" state.
Going to fix ASAP.

python/_brotli.cc
```diff
@@ -88,6 +87,37 @@ static int lgblock_convertor(PyObject *o, int *lgblock) {
   return 1;
 }
 
+static BROTLI_BOOL compress_stream(BrotliEncoderState* enc, BrotliEncoderOperation op,
+                                   std::vector<uint8_t>* output, uint8_t* input, size_t input_length) {
+  BROTLI_BOOL ok = BROTLI_TRUE;
+
+  while (ok) {
+    size_t available_in = input_length;
+    const uint8_t* next_in = input;
```
eustas: available_in and next_in are reset at the beginning of the loop.
This will cause almost infinite loop, if output does not fit internal encoder buffer...

nicksay: Ah, silly mistake, thanks for catching.  I moved the variable outside the loop in the next (temporary) commit.

python/_brotli.cc
nicksay: I'll sync up once #451 is merged.

nicksay:
_updated commit description follows_

---

This adds `flush` and `finish` methods to the `Compressor`
object in the extension module, renames the `compress` method to
`process`, and updates it to only process data. Now, one or more
`process` calls followed by a `finish` call will be equivalent
to a module-level `compress` call.

Note: To maximize the compression efficiency (and match
underlying Brotli behavior, the `Compressor` object `process`
method does not guarantee all input is immediately written to
output. To ensure immediate output, call `flush` to manually
flush the compression buffer. Extraneous flushing can increase
the size, but may be required when processing streaming data.

Progress on #191

python/_brotli.cc
```diff
@@ -88,6 +87,38 @@ static int lgblock_convertor(PyObject *o, int *lgblock) {
...
+                                   std::vector<uint8_t>* output, uint8_t* input, size_t input_length) {
+  BROTLI_BOOL ok = BROTLI_TRUE;
+
+  size_t available_in = input_length;
+  const uint8_t* next_in = input;
+  size_t available_out = 0;
+  uint8_t* next_out = NULL;
+
+  while (ok) {
+    ok = BrotliEncoderCompressStream(enc, op,
```
nicksay: @eustas The `available_in` and `next_in` variables are now outside of the loop.  In my testing, if `available_out` is `0`, then it won't be modified by `BrotliEncoderCompressStream`.  Would you prefer that I reset `available_out` to `0` each call?

python/_brotli.cc
```diff
@@ -88,6 +87,38 @@ static int lgblock_convertor(PyObject *o, int *lgblock) {
...
+                                     &available_in, &next_in,
+                                     &available_out, &next_out, NULL);
+    if (!ok)
+      break;
+
+    size_t buffer_length = 0; // Request all available output.
+    const uint8_t* buffer = BrotliEncoderTakeOutput(enc, &buffer_length);
```
nicksay: This is working now after #451, thanks.

python/_brotli.cc
```diff
@@ -243,7 +343,9 @@ static PyMemberDef brotli_Compressor_members[] = {
 };
 
 static PyMethodDef brotli_Compressor_methods[] = {
-  {"compress", (PyCFunction)brotli_Compressor_compress, METH_VARARGS, brotli_Compressor_compress_doc},
+  {"process", (PyCFunction)brotli_Compressor_process, METH_VARARGS, brotli_Compressor_process_doc},
```
nicksay: FYI I went with the `process`, `flush`, `finish` naming to match the underly Brotli behavior.

nicksay:
(Rebased to fix a bug in the `compress_stream` loop; it seems I have to carry comments forward.)

nicksay:
(Actually, no; I see the comments survived the rebase, but you have to [view latest changes](https://github.com/google/brotli/pull/448/files) instead of clicking the previous direct links.)

nicksay:
@eustas looks like the Travis osx gcc 4.4 config still isn't working: https://travis-ci.org/google/brotli/jobs/169758805

nicksay:
@eustas Nevermind! The Travis build was just flaky, and everything is passing again: https://travis-ci.org/google/brotli/builds/170025117

python/_brotli.cc
eustas: According to contract, it should not be changed -> no need to reassign it =)

PR #450, opened 2016-10-18
Build shared libraries by default

eustas:
* Declare `BUILD_SHARED_LIBS` option for CMake
* Define `${LIB}_SHARED_COMPILATION` when compiling shared library
* Define and use `BROTLI_xxx_API`
* Fix remaining unprefixed defines in port.h


eustas:
So, now we have a diversity: CMake builds shared libraries by default and Python build static libraries (by design?)

PR #458, opened 2016-10-25
Python: Add tests for streamed compression

nicksay:
Progress on #191

### DIFF
--- a/python/tests/compressor_test.py
+++ b/python/tests/compressor_test.py
@@ -1,0 +1,141 @@
+# Copyright 2016 The Brotli Authors. All rights reserved.
+#
+# Distributed under MIT license.
+# See file LICENSE for detail or copy at https://opensource.org/licenses/MIT
+
+import filecmp
+import functools
+import os
+import sys
+import types
+import unittest
+
+import test_utils
+
+import brotli
+
+
+TEST_DATA_FILES = [
+    'empty',  # Empty file
+    '10x10y',  # Small text
+    'alice29.txt',  # Large text
+    'random_org_10k.bin',  # Small data
+    'mapsdatazrh',  # Large data
+]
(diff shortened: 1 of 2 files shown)


anthrotype:
the appveyor builds failure seems to be related to the fact that PowerShell sometimes reports a `NativeCommandError` when a process writes to STDERR (in our case, it seems to be the unittest module)

E.g. see http://stackoverflow.com/a/22615314

some users suggest to redirect stderr to null as a workaround

```diff
diff --git a/appveyor.yml b/appveyor.yml
index bcabe41..268142b 100644
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -88,7 +88,7 @@ test_script:
       if ($env:BUILD_SYSTEM -eq "CMake") {
         ctest --output-on-failure --interactive-debug-mode 0 -C Debug
       } else {
-        python setup.py build test
+        python setup.py build test 2> $null
       }

 after_test:
```

Note that the custom `TestCommand` in setup.py was meant to emulate with python the  compatibility/roundtrip shell scripts written for the `bro` command line tool.
There are better ways to discover Python test modules, e.g. with setuptools [`test_suite`](http://setuptools.readthedocs.io/en/latest/setuptools.html?highlight=test_suite#test-build-package-and-run-a-unittest-suite) keyword, or with `pytest` (which also supports discovering `unittest`-style modules).

nicksay:
I've pulled out the `s/build_ext/build/` change into #460 and I'm debugging the appveyor bug in an isolated branch now.  I'm planning on moving to `unittest`-style discovery for tests in the future.

nicksay:
I've fixed the Appveyor issue in #461.  I created a temp branch that combines all 3 PRs and you can see the new tests passing at https://ci.appveyor.com/project/nicksay/brotli/build/1.0.12 and https://travis-ci.org/nicksay/brotli/builds/170921320.  Once #460 and #461 are merged, I'll update this PR and everything should be good to go.

PR #460, opened 2016-10-26
Python: Use "build" instead of "build_ext" in scripts

nicksay:
---

_edited Oct 28: see updated commit description and in the thread below_

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -39,7 +39,7 @@ case "$1" in
 		if [ "${TRAVIS_OS_NAME}" = "osx" ]; then
 			source venv/bin/activate
 		fi
-		python setup.py build_ext test
+		python setup.py build test
 		;;
 	esac
 	;;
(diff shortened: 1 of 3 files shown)


eustas:
Hello. What is the difference between "build" and "build_ext"?

eustas:
(ooops, rebase is required)

anthrotype:
build will also compile the pure python modules, besides in turn calling build_ext to compile the extension module. It is the correct command to use, now that brotli is a package with some pure python modules. Previously it was a single extension module.

nicksay:
_updated commit description follows_

---

Previously, the Python package consisted of a single extension
module, so `build_ext` was sufficient.  Now, the package
contains a native module and an extension module, so both
`build_py` and `build_ext` are required.  Instead, run `build`,
which calls both `build_py` and `build_ext` automatically.

nicksay:
Also, I rebased to the latest commit on `master`

PR #464, opened 2016-10-31
FInishing touches for installing libbrotli with CMake

### LINKED ISSUE #326: libbrotli

nemequ:
AFAIK this should fix #326.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,3 +263,128 @@ if(NOT BROTLI_DISABLE_TESTS)
         -P ${CMAKE_CURRENT_SOURCE_DIR}/tests/run-compatibility-test.cmake)
   endforeach()
 endif()
+
+# Generate a pkg-config file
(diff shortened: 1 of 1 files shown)


eustas:
Thank you! 👍

PR #471, opened 2016-12-02
Fix build with -Wconditional-uninitialized.

PiotrSikora:
This is false positive, since it will always get set before use, but clang doesn't like it...


eustas:
Thanks.

PR #479, opened 2016-12-09
Update python brotli wrapper

### LINKED ISSUE #476: Release the Python GIL when compressing/decompressing
The C extension doesn't currently release the GIL during [`compress_stream()`](https://github.com/google/brotli/blob/master/python/_brotli.cc#L92) and during `decompress_stream()` (between https://github.com/google/brotli/blob/master/python/_brotli.cc#L433 and https://github.com/google/brotli/blob/master/python/_brotli.cc#L452 which is a shame because we can enable greater parallelism .

eustas:
* release GIL on CPU intensive blocks, fixes #476
 * use BrotliDecoderTakeOutput (less memory, less memcpy)


eustas:
@nicksay, could you review, please.

nicksay:
Changes look good, but https://travis-ci.org/google/brotli/jobs/182575377 failed for some reason.  Maybe rebase on master to get the new unit tests to double-check?

eustas:
Finished well after rerun. Going to rebase and retest as a whole.

Issue #473, opened 2016-12-03
Split pc file into encoding and decoding libraries


fred-wang:
A woff2 generator or consumer will generally use only one of this library so it's better to have separate pc files. This is how it was done in @bagder 's initial commit but @nemequ only created a single pc file.

There were discussions about the names to use for pc files, but I'm not sure how important it is.

fred-wang:
@eustas Are you going to take care of this?

eustas:
Surely. Will try to do it today.

eustas:
Please take a look at #490

fred-wang:
Thanks. I trust you @eustas !

PR #490, opened 2016-12-21
Split .pc files (#473)

eustas:
* split brotli common/dec/enc .pc files
* add URL, and use DEPENS_PRIVATE generator params

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -313,8 +313,8 @@ endfunction(generate_pkg_config_path)
 
 function(generate_pkg_config output_file)
   set (options)
-  set (oneValueArgs NAME DESCRIPTION PREFIX LIBDIR INCLUDEDIR VERSION)
-  set (multiValueArgs DEPENDS CFLAGS LIBRARIES)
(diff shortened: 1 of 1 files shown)


Earnestly:
Why was this done?

Everything once configured to use `pkg-config ... brotli` now fails...

eustas:
We are terribly sorry for the inconveniences. It was a difficult decision that hit a lot of fans of brotli. But it has to be done to cover more use cases.
In some use-cases brotli encoder is an unbearable burden. So, we decided to split brotli into `encoder` and `decoder` modules. `common` is a part required by both of them; also, there is an alternative variant of `common` module that does not bring dictionary and allows loading it (for very special use-cases).

Earnestly:
I'm not entirely sure why this would require the removal of `brotli.pc` though.  I can understand why you might want to split it for use cases which don't require linking everything, but why remove the `brotli.pc`?

eustas:
It is an unfortunate consequence: we try to avoid redundancy.

Earnestly:
I see, that's an unfortunate view to hold but at least there is a rational for others who might encounter this issue as systems are updated.

PR #495, opened 2017-01-25
Only use __builtin_clz on GCC >= 3.4

mmuman:
Haiku at least still requires GCC 2 for system components.


mmuman:
Ugh, why would I want to bring my gmail account into this? For a oneliner!?

eustas:
Oh, this ifdef wasn't retouched for almost 3 years.
Going to fix it with definitions taken from port.h very soon.

eustas:
Will be fixed with #498

mmuman:
thx!

PR #496, opened 2017-01-25
Haiku: force undefine INFINITY for GCC2 as it's not const

mmuman:
It's defined as a non-const inline union...
cf. [our current math.h](http://cgit.haiku-os.org/haiku/tree/headers/posix/math.h#n68)
I'm open to a cleaner solution. Maybe we should make it const, not sure though...

### DIFF
--- a/enc/backward_references.c
+++ b/enc/backward_references.c
@@ -26,6 +26,10 @@
 extern "C" {
 #endif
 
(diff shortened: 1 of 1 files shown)


eustas:
Fixed with #497

mmuman:
Thank!

PR #498, opened 2017-01-26
Encoder: modernize builtin clz guard

eustas:


eustas:
Going to push bigger update very soon.

PR #505, opened 2017-02-07
Fix #502 decoder bug

### LINKED ISSUE #502: Non-compliant decoding according to section 4?
I was running fuzz testing on [my Brotli implementation](https://github.com/dsnet/compress/tree/master/brotli) and noticed that the following input differs when decompressed by my version compared to the canonical one here.

The input causing issues (in hex): `e2e184886a563080e0391603e430f9`

When decompressed by the `google/brotli` C implementation: 
`worldwhitemediafirstirstirstirstirstirstirstirstirstirstirst...`
When decompressed by my implementation: 
`worldorldorldorldorldorldorldorldorldorldorldorldorldorldorl...`

According to section 4 of the RFC:
>   When a distance symbol 0 appears, the distance it represents (i.e.,
   the last distance in the sequence of distances) is not pushed to the
   ring buffer of last distances

However, this seems contrary to [what C decoder currently does](https://github.com/google/brotli/blob/ccabf811ffcef087d9f10cf417b05271c2067a82/dec/decode.c#L1706-L1710):
```c
if (s->distance_code >= 0) {
	--s->dist_rb_idx;
	s->distance_code = s->dist_rb[s->dist_rb_idx & 3];
	goto postReadDistance;  /* We already have the implicit distance */
}
```

The logic in the C version seems contrary to the stated prose in the RFC, which makes no mention that an implicit zero *reverse* rotates the distance ring-buffer.

eustas:
Decoder may have produced invalid output if:
 * at offset 0..3 dictionary word with index 3..0 for some length N is used
   and distance is encoded with direct distance code 0, and
 * at least one of next 4 commands use value from distance ringbuffer

### DIFF
--- a/dec/decode.c
+++ b/dec/decode.c
@@ -1391,6 +1391,8 @@ static BROTLI_INLINE void TakeDistanceFromRingBuffer(BrotliDecoderState* s) {
   if (s->distance_code == 0) {
     --s->dist_rb_idx;
     s->distance_code = s->dist_rb[s->dist_rb_idx & 3];
+    /* Compensate double distance-ring-buffer roll for dictionary items. */
+    s->distance_context = 1;
   } else {
     int distance_code = s->distance_code << 1;
     /* kDistanceShortCodeIndexOffset has 2-bit values from LSB: */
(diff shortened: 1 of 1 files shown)


eustas:
FYI: encoder does not use short codes (distance ring-buffer) for static dictionary entries -> all streams produced by encoder should be decodable even without fix; this fix provides better compatibility with future and third party encoders.

PR #511, opened 2017-02-17
Blacklist PGI from using conformant array parameters.

nemequ:
There is a bug in pgcc with conformant array parameters where the
length argument is a pointer which triggers a compiler error
(PGC-S-0094, to be specific).  The issue has been reported to PGI and
is being tracked internally as TPR 23778.  For more information, see

  https://www.pgroup.com/userforum/viewtopic.php?t=5501

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -21,6 +21,13 @@ case "$1" in
 			;;
 		esac
 		;;
+	    "linux")
+		case "${CC}" in
+		    "pgcc")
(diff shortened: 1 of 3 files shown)


nemequ:
If you're interested, I can also add a test for PGI on Travis (via [nemequ/pgi-travis](https://github.com/nemequ/pgi-travis)).

eustas:
Hi. Thank you for help. Yup, PGI environment would be nice to have.

nemequ:
PR updated w/ a PGI build.

eustas:
Thank you.

PR #518, opened 2017-03-01
Fix CMakeLists.txt specifying a nonexistent pkgconfig package

iand675:
The pkgconfig package is named libbrotlicommon, not brotlicommon. 

This makes the generated pkgconfig package for libbrotlienc actually work.


eustas:
👍

PR #523, opened 2017-03-20
Better compression

eustas:
* use more complex context modelling on 1MiB+ files


go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Distributed under MIT license.
+// See file LICENSE for detail or copy at https://opensource.org/licenses/MIT
+
+// Package cbrotli compresses and decompresses data with C-Brotli library.
+package cbrotli
```
dsnet: The convention in the Go community is to name the parent directory the same as package name.

https://golang.org/doc/effective_go.html#package-names says:
> Another convention is that the package name is the base name of its source directory; the package in src/encoding/base64 is imported as "encoding/base64" but has name base64, not encoding_base64 and not encodingBase64.

Thus, creating a directory `go/cbrotli` and placing this file in `go/cbrotli/cbrotli.go`. It is a redundant when there is only 1 package, but as a convention, it scales better.

Bazel allows for both the old-style and current representation (because Bazel supports the package layout that preceded the design of today's conventional layout), but this is not the norm for most Go users.

go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
...
+
+//------------------------------------------------------------------------------
+// Encoder
+//------------------------------------------------------------------------------
+
+// WriterOptions configures Writer.
+type WriterOptions struct {
+	// Quality controls the compression-speed vs compression-density trade-offs.
+	// The higher the quality, the slower the compression. Range is 0 to 11.
+	Quality int
```
dsnet: Comment that 0 is effectively: "no compression"?

go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
...
+	// LGWin is the base 2 logarithm of the sliding window size.
+	// Range is 10 to 24. 0 indicates automatic configuration based on Quality.
+	LGWin int
+}
+
+// ErrEncodeFailure is returned when the encode operation fails.
+var ErrEncodeFailure = errors.New("Brotli encode failure")
+
+// ErrWriteAfterClose is returned from Write() if called after Close().
+var ErrWriteAfterClose = errors.New("brotli.Writer: write after close")
```
dsnet: Unexport this. I know the [`archive/tar.ErrWriteAfterClose`](https://golang.org/pkg/archive/tar/#pkg-variables) does this, but I consider that a mistake in retrospect.

go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
...
+	// Quality controls the compression-speed vs compression-density trade-offs.
+	// The higher the quality, the slower the compression. Range is 0 to 11.
+	Quality int
+	// LGWin is the base 2 logarithm of the sliding window size.
+	// Range is 10 to 24. 0 indicates automatic configuration based on Quality.
+	LGWin int
+}
+
+// ErrEncodeFailure is returned when the encode operation fails.
+var ErrEncodeFailure = errors.New("Brotli encode failure")
```
dsnet: We should not export these errors. Singleton error values are generally not a good design pattern. 

My recommendation is to not export these and add them later if someone needs them. Once added, the API is hard to change once users are depending on it.

go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
...
+// NewWriter initializes new Writer instance.
+// Close MUST be called to free resources.
+func NewWriter(dst io.Writer, options WriterOptions) *Writer {
+	return &Writer{
+		dst:     dst,
+		encoder: encoder.New(options.Quality, options.LGWin),
+	}
+}
+
+// Close implements io.Closer. Close() MUST be invoked to free native resources.
```
dsnet: Nit: Go style typically avoids the `()` when referring to functions.

E.g., `Close MUST be invoked to free native resources.`

go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+func (z *Writer) writeChunk(p []byte, op encoder.Operation) (int, error) {
+	if z.closed {
+		return 0, ErrWriteAfterClose
+	}
+	var totalBytesConsumed int
+	var err error
+	totalBytesConsumed = 0
+	err = nil
```
dsnet: This is not necessary. 

`var totalBytesConsumed int` and `var err error` implies `totalBytesConsumed == 0 && err == nil`

go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
...
+
+//------------------------------------------------------------------------------
+// Decoder
+//------------------------------------------------------------------------------
+
+// ErrDecodeFailure is returned the decode operation fails.
+var ErrDecodeFailure = errors.New("Brotli decode failure")
+
+// ErrReadAfterClose indicates Read has been called after Close.
+var ErrReadAfterClose = errors.New("Read() after Close()")
```
dsnet: Unexport these errors.

go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
...
+		z.decoder.Close()
+		z.err = ErrReadAfterClose
+	}
+	return nil
+}
+
+func isEOF(src io.Reader) bool {
+	var buf []byte
+	buf = make([]byte, 1)
+	n, err := src.Read(buf)
```
dsnet: Perhaps `src.Read(make([]byte, 1))` and avoid the above two lines?

go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
...
+	if z.err != nil {
+		return 0, z.err
+	}
+	// See io.Reader documentation.
+	if len(p) == 0 {
+		return 0, nil
+	}
+
+	var totalOutBytes int
+	totalOutBytes = 0
```
dsnet: Unnecessary.

go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
...
+
+	// There is no practical limit for amount of bytes being consumed by decoder
+	// before producing any output. Continue feeding decoder until some data is
+	// produced
+	for {
+		// Push already produced output first.
+		var outBytes int
+		outBytes = len(z.output)
```
dsnet: `outBytes := len(z.output)` to avoid `var outBytes int`

dsnet: Furthermore, I would put the declaration of `outBytes` in the if statement to make it clear that the scope of this variable is only in the following if statement.

For example:
```go
if outBytes := len(z.output); outBytes > 0 {
	...
}
```

go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
...
+		if outBytes != 0 {
+			outBytes = copy(p, z.output)
+			p = p[outBytes:]
+			z.output = z.output[outBytes:]
+			totalOutBytes += outBytes
+			// Output buffer is full.
+			if len(p) == 0 {
+				break
+			}
+			continue
```
dsnet: IIUC, hitting this line implies that `len(z.output) == 0` since all of it would have been copied to `p` otherwise? Thus, should we just drop this `continue` and move on in the loop?

go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
...
+		// Do decoding.
+		consumed, output, status := z.decoder.DecompressStream(z.next)
+		z.output = output
+		z.next = z.next[consumed:]
+		if status == decoder.Error {
+			// When error happens, the remaining output does not matter.
+			z.err = ErrDecodeFailure
+			break
+		} else if status == decoder.Done {
+			// Decoder stream is closed; not further input is expected.
```
dsnet: Nit: s/not further error/no further error/

go/cbrotli.go
```diff
@@ -0,0 +1,264 @@
...
+			}
+			// No more output is expected; keep pushing output.
+			z.eof = true
+			continue
+		} else {
+			// If can not move any further...
+			if consumed == 0 && len(z.output) == 0 {
+				// Unexpected end of input.
+				if z.srcErr == io.EOF || totalOutBytes == 0 {
+					z.err = ErrDecodeFailure
```
dsnet: Return `io.ErrUnexpectedEOF` instead.

go/cgo/decoder.go
```diff
@@ -0,0 +1,110 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Distributed under MIT license.
+// See file LICENSE for detail or copy at https://opensource.org/licenses/MIT
+
+// Package decoder wraps the brotli decoder C API used by package brotli.
+package decoder
```
dsnet: This package and `encoder` seem like it is an internal implementation detail to cbrotli. Perhaps hide this under an internal package directory?

https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit

go/cbrotli.go
eustas: Fair point

go/cbrotli.go
eustas: Done/

go/cbrotli.go
eustas: Done.

go/cbrotli.go
eustas: Fixed.

go/cbrotli.go
eustas: plain C habit =)

go/cbrotli.go
eustas: Done

go/cbrotli.go
eustas: Nice!

go/cbrotli.go
eustas: Fixed

go/cbrotli.go
eustas: :+1:

go/cbrotli.go
eustas: You are right. But this "continue" is added for readability, so workflow (without reading details) is clear: "push while we have something to push".

go/cbrotli.go
eustas: Thanks.

go/cbrotli.go
eustas: Done. But isn't it a singleton error variable?

eustas:
Rolled back go-wrapper for now. Going to add it in next PR.

PR #533, opened 2017-03-29
cbrotli: simplify Go implementation and reduce copying

bcmills:
For Read and Write calls, we can pass the Go-allocated slice buffers
directly to the C API. (They do not contain Go pointers, so this
cross-language aliasing is allowed for the duration of the call.)

io.Reader and io.Writer are the Go primitives for this kind of I/O, so
implement those directly instead of providing separate internal
encoder and decoder APIs. (The internal packages weren't laid out to
be compatible with the go toolchain anyway, making this wrapper
awkward use with `go get` instead of Bazel.)


bcmills:
(Attn: @eustas @dsnet)

go/cbrotli/writer.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	for {
+		var written C.size_t
+		ok := C.CompressStream(w.state, op,
+			(*C.uint8_t)(&w.buf[0]), C.size_t(len(w.buf)), nil, 0, &written, nil)
+		w.encoded = w.buf[:int(written)]
+		if err := w.flushBuf(); err != nil {
+			return err
+		}
+		if written == 0 {
```
eustas: `written == 0` doesn't mean that all input is consumed. This might mean that encoder can't take more input without producing output.

bcmills: This is the flush / finish path: there is no input on this call.  (What am I missing?)

eustas:
https://github.com/bcmills/brotli/tree/cbrotli/go/cbrotli

eustas:
Going to be pushed with next update.

PR #541, opened 2017-04-21
Move files & update paths

eustas:
Cleanup root directory:
 * move CI files into `scripts/`
 * move `common`, `dec`, `enc`, `include`, `tools`, `fuzz` into `c/`


configure-cmake
```diff
@@ -32,8 +32,8 @@ PREFIX=/usr/local
 LIBDIR=
 CMAKE_ARGS=
 
-if [ -e "${TOP_SRCDIR}/.configure-custom.sh" ]; then
-    . "${TOP_SRCDIR}/.configure-custom.sh"
+if [ -e "${TOP_SRCDIR}/script/.configure-custom.sh" ]; then
```
szabadka: This should be 'scripts'.

eustas: Nice catch! Fixed.
Thanks!

PR #545, opened 2017-05-04
turn java library into an OSGi bundle

bodewig:
This makes the Maven build turn the created jar into an OSGi bundle. I'm not sure whether Maven or Bazel is the preferred build tool, though.

### DIFF
--- a/java/org/brotli/dec/pom.xml
+++ b/java/org/brotli/dec/pom.xml
@@ -13,6 +13,11 @@
 
   <name>${project.groupId}:${project.artifactId}</name>
 
+  <properties>
+    <manifestdir>${project.build.directory}/osgi</manifestdir>
+    <manifestfile>${manifestdir}/MANIFEST.MF</manifestfile>
+  </properties>
+
   <build>
     <sourceDirectory>../../..</sourceDirectory>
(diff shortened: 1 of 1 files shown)


eustas:
Grüße, Stefan. Thanks for the PR.

Is it a standalone change, or releasing of new artifact version is required?

bodewig:
Thanks Eugene!

If you want to make downstream OSGi users happy, they'll need a new artifact.

We've just merged a pull request for Apache Commons Compress that adds read-only support for brotli based on dec and all Apache Commons projects try to be bundles. We've marked the dependency as optional so people using Compress aren't directly affected. We haven't cut a release with it, yet, but I think we are close. If you are interested in publishing new artifacts, I'd be happy to hold back the release and upgrade the dependency.

eustas:
Going to push a little update to decoder soon that provides 5-25% speedup and publish new artifact right after that. Hopefully, this will happen tomorrow.

bodewig:
Sounds great, thanks!

eustas:
Just released 0.1.2

bodewig:
Wow, thanks a lot.

https://github.com/apache/commons-compress/commit/0eccda844e59b6d5bdd06e62710bbf09dacc040d

PR #546, opened 2017-05-05
Python: Decompressor: Streaming decompression support

ttnbs:
python-brotli has Compressor for streaming compression but nothing for
streaming decompression.
This is a straight-forward copy of the Compressor code into the new
class Decompressor.

### DIFF
--- a/python/_brotli.cc
+++ b/python/_brotli.cc
@@ -401,6 +401,241 @@ static PyTypeObject brotli_CompressorType = {
   brotli_Compressor_new,                 /* tp_new */
 };
 
+static BROTLI_BOOL decompress_stream(BrotliDecoderState* dec,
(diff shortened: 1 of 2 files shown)


ttnbs:
I signed it!

eustas:
Hello. Thanks for the PR!
Just returned from vacation. Going to review your PR this week.

eustas:
Hi. Sorry for the super-long review. Will get back to you as soon as get results from internal reviewer.

nicksay:
Apologies (again) for the delay here.

The extension implementation looks right to me, but we should have some tests to verify correctness before exposing the Python Decompressor class in the module.  To keep from blocking this any longer, @eustas will add a simple test after landing this PR.

PR #557, opened 2017-06-06
Readme improvements

Chocobo1:
* Use tools.ietf.org for displaying RFC7932
  tools.ietf.org has HTML links which is helpful when reading in browser
* Add appveyor badge


eustas:
Thanks.

PR #562, opened 2017-06-16
Fix compilation issue with BROTLI_ALLOC macro using GCC 7

### LINKED ISSUE #561: Compilation error due to macro BROTLI_ALLOC in enc/memory.h (GCC 7 -Wint-in-bool-context)
Hi,

I'm compiling ngx_brotli, with v0.6.0 brotli release (but the macro is the same in current master branch), and I had to fix an issue with this macro:
```
../ngx_brotli_test/deps/brotli/enc/./block_splitter_inc.h:403:61: error: ‘*’ in boolean context, suggest ‘&&’ instead [-Werror=int-in-bool-context]
     double* insert_cost = BROTLI_ALLOC(m, double, data_size * num_histograms);
../ngx_brotli_test/deps/brotli/enc/././memory.h:43:5: note: in definition of macro ‘BROTLI_ALLOC’
   ((N) ? ((T*)BrotliAllocate((M), (N) * sizeof(T))) : NULL)
     ^
```
Just replace the macro like this:
```c
#define BROTLI_ALLOC(M, T, N)                               \
  ((N) > 0 ? ((T*)BrotliAllocate((M), (N) * sizeof(T))) : NULL)
```
(test if (N) is strictly greater that zero).

I believe this is the fix, it compiles and runs fine after changing this, but only tried using nginx with ngx_brotli (it does not use the decoder part).

I'm using GCC 7.1.1 (as available in ArchLinux repos); nginx does **not** add a parameter like `-std` to commands.

```
cc -c -pipe  -O -W -Wall -Wpointer-arith -Wno-unused-parameter -Werror -g -fPIE -D_FORTIFY_SOURCE=2 -fstack-check -fstack-protector-strong -Wno-deprecated-declarations  -I src/core -I src/event -I src/event/modules -I src/os/unix -I objs -I src/http -I src/http/modules -I ../ngx_brotli_test/deps/brotli/include \
	-o objs/addon/enc/block_splitter.o \
	../ngx_brotli_test/deps/brotli/enc/block_splitter.c
```

ngx_brotli repo actually uses the commit 222564a95d9ab58865a096b8d9f7324ea5f2e03e for brotli, if there is any security concern about this.

If you agree to, I can try to submit a PR (on both repos, in ngx_brotli it's just updating the list of files in enc) so I can have my name somewhere in the history :)

Exagone313:
Fixes #561.
The issue is with GCC 7 that adds a check `-Wint-in-bool-context`.
Maybe you'd prefer to disable this check instead?

(Note: I haven't signed the CLA yet.)

### DIFF
--- a/c/enc/memory.h
+++ b/c/enc/memory.h
@@ -40,7 +40,7 @@ BROTLI_INTERNAL void BrotliInitMemoryManager(
 
 BROTLI_INTERNAL void* BrotliAllocate(MemoryManager* m, size_t n);
 #define BROTLI_ALLOC(M, T, N)                               \
-  ((N) ? ((T*)BrotliAllocate((M), (N) * sizeof(T))) : NULL)
+  ((N) > 0 ? ((T*)BrotliAllocate((M), (N) * sizeof(T))) : NULL)
 
 BROTLI_INTERNAL void BrotliFree(MemoryManager* m, void* p);
 #define BROTLI_FREE(M, P) { \


Exagone313:
I signed it!

eustas:
The fix looks good. I think it is better to have slightly more explicit checks than loose a chance, when compiler finds a real issue =)

eustas:
Going to pick the fix to v0.6 branch

PR #567, opened 2017-06-24
Added windows platform support to premake

tavrez:
Win32 and Win64 configuration support for visual studio solutions

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -1,7 +1,7 @@
 -- A solution contains projects, and defines the available configurations
 solution "brotli"
 configurations { "Release", "Debug" }
-platforms { "Win32", "Win64"}
+platforms { "x64", "x86" }
 targetdir "bin"
 location "buildfiles"
 flags "RelativeLinks"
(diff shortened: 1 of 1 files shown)


tavrez:
I've signed CLA

eustas:
Sorry for the late response. Lets change the order in the platforms list, so Win64 is listed first (and becomes the default).

eustas:
Also adding "platform" blows the configuration matrix.
Now for linux/gmake there will be:
 * release_win32
 * release_win64
 * debug_win32
 * debug_win64

Sounds ridiculous. So, lets get rid of "win" prefix.

eustas:
There is a nice example how to specify architecture on premake wiki: https://github.com/premake/premake-core/wiki/architecture

tavrez:
I though that `system:"Windows"` in filter configurations will make it only available in windows and results won't change anything in linux/gmake.
Thanks for the response and notes, I'll make it better

premake5.lua
```diff
@@ -13,6 +14,12 @@ filter "configurations:Release"
 filter "configurations:Debug"
   flags { "Symbols" }
 
+filter { "platforms:x64" }
+   architecture "x64"
```
eustas: x86?

premake5.lua
```diff
@@ -13,6 +14,12 @@ filter "configurations:Release"
...
+
+filter { "platforms:x86" }
+   architecture "x86_64"
```
eustas: x86?

tavrez:
Sorry, seems like VS only care about platform name, not what we write in front of architecture. But I fixed mistyping you mentioned.

PR #572, opened 2017-07-10
Export functions on win platform

Vedin:
These decode methods not currently exported on win platforms without BROTLI_DEC_API modifier.  
They are currently being exported for OSX and Linux platforms.


eustas:
Thanks!

PR #578, opened 2017-08-02
Python: Add a "make install" command and clarify documentation

### LINKED ISSUE #576: Command "python setup.py install" not work :-(
python setup.py install 
not work. error:
C:\DevTools\Python\python.exe: can't open file 'setup.py': [Errno 2] No such file or directory

nicksay:
Closes #576

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,12 +66,12 @@ See [Premake5](https://premake.github.io/)
 
 #### Python
 
-To install the Python module from source, run the following:
+To install the latest release of the Python module, run the following:
 
-    $ python setup.py install
+    $ pip install brotli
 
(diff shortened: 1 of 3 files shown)


anthrotype:
May I suggest that you also mention here in the Readme that the Brotli python module is also available on the Python Package Index (PyPI), precompiled for all cpython versions and the three major desktop platforms, and the way to install is `pip install brotli`?

nicksay:
Good idea

nicksay:
@anthrotype Updated both `README.md` and `python/README.md` to reference `pip install brotli`

anthrotype:
LGTM thanks

PR #579, opened 2017-08-02
Add an EditorConfig file to provide consistent style across editors.

nicksay:


eustas:
Could this reside not in root directory?

nicksay:
We don't have to have it all, but I think it works best in the root directory so that the settings apply to all languages. Up to you!

PR #583, opened 2017-08-08
disable buidling/deployment of python wheels

anthrotype:
Since this is now done in a separate https://github.com/google/brotli-wheels, it makes sense remove that from the upstream brotli's Travis and Appveyor setup.
We still run `python setup.py test` on both 2.7 and 3.6 on linux and windows, as well as on the system's python on macOS.


anthrotype:
I noticed that we were re-building the extension module three times before actually running the tests...

It turns out

1) the code that checks if the extension module was newer than its source files was not there in our implementation of `build_ext.build_extension` method (I just copied it from cpython's [distutils](https://github.com/python/cpython/blob/v3.6.2/Lib/distutils/command/build_ext.py#L485-L500))

2) there was a typo in the list of headers (a missing comma) which made it so that two headers' paths were combined into one, but the latter did not exist and thus distutils/setuptools would think the extension module was not up-to-date (as one of the dependencies was showing as missing), and would always run the `built_ext` command over and over.

anthrotype:
ok.. We went from running `build_ext` three times down to running it twice...
That's because running the `test` command will automatically run `build_ext` as a dependent subcommand, so we don't need `build` and `test` in a single setup.py invocation.

With the last commit we call the compiler only once (384f8b6), by simply doing `python setup.py test`.

anthrotype:
linux python 2.7 build went down from 3m 40s to 2m 12s. nice

setup.py
```diff
@@ -53,6 +55,21 @@ def get_source_files(self):
         return filenames
 
     def build_extension(self, ext):
+        if ext.sources is None or not isinstance(ext.sources, (list, tuple)):
+            from distutils.errors import DistutilsSetupError
```
nicksay: I think it's okay to move this import to the top

setup.py
```diff
@@ -15,6 +15,8 @@
     from distutils.core import Extension
     from distutils.core import setup
 from distutils.command.build_ext import build_ext
+from distutils.dep_util import newer_group
```
nicksay: Let's try to import packages/modules instead of functions:
https://google.github.io/styleguide/pyguide.html#Imports

setup.py
```diff
@@ -53,6 +55,21 @@ def get_source_files(self):
...
+        if ext.sources is None or not isinstance(ext.sources, (list, tuple)):
+            from distutils.errors import DistutilsSetupError
+            raise DistutilsSetupError(
+                "in 'ext_modules' option (extension '%s'), "
+                "'sources' must be present and must be "
+                "a list of source filenames" % ext.name)
+
+        ext_path = self.get_ext_fullpath(ext.name)
+        depends = ext.sources + ext.depends
+        if not (self.force or newer_group(depends, ext_path, 'newer')):
```
nicksay: What are your thoughts on `"newer"` vs  the default `"error"`?

python/README.md
```diff
@@ -24,7 +24,7 @@ able to edit the source files.
 
 For convenience, you may run the following commands from this directory:
 
-    $ make          # Deploy the module in "development mode"
+    $ make          # Build the module in-place
```
nicksay: Maybe leave something about "development mode" here since it's described in the paragraph right before this?

anthrotype:
@nicksay FYI I have modified the Makefile in 44b5e66 so that the default `make` target does not call the `develop` command, which is I believe is too intrusive for a simple default build target.
It's good for developers who want to actually modify the code, but for the sake of testing the build it's better to not modify the user's python environment (e.g. say one does not have admin rights, `develop` command will fail, just like `install`).

Also, we don't need to tell `make` about the dependency between 'test' and 'build' targets as that is baked in the `python setup.py test` command as I explained above.

python/README.md
anthrotype: oh I see now, sorry.. I can actually add an additional `make develop`, wdyt?

nicksay: Yeah, that sounds good to me, though perhaps we should remove the references to "development mode" in the README if the in-place build is the default.

setup.py
anthrotype: don't know.. I just copied from the default distutils `build_ext` module. I thought it'd be better to keep the default behavior. But I see that with "error" we could have avoided issues like the missing comma..

anthrotype:
@nicksay I added the `make develop` target and reworked the python/README.md so that the reference to setuptools development mode is still there but no longer at the top of the section.

I noticed also that the `test` command is setuptools only, and the brotli setup.py still works with plain distutils (setuptools is widely available in modern python distributions but is not strictly part of the standard library). So I decided to revert the `make test` command to calling `python -m unittest discover...` as it was before, for backward compatiblity's sake.

anthrotype:
The last Travis failure is just a random and unrelated, this is good to go now.

nicksay:
Still LGTM 👍

svnpenn:
https://github.com/google/brotli/pull/583

svnpenn:
https://github.com/google/brotli/issues/591

PR #587, opened 2017-08-21
Also build test with GCC 7 and LLVM/Clang 4.0

paulmenzel:


paulmenzel:
I don’t have a Google work account. How can I sign the CLA?

eustas:
Hello, Paul.

 Unfortunately, I haven't found any information on how non-gmail users can sign CLA. Perhaps, one possible way could be "Corparate CLA", but it seems to be more difficult...

eustas:
Included into ##598

PR #618, opened 2017-10-26
Support MSVC intrinsic in Log2FloorNonZero()

KindDragon:


eustas:
Hello.

 Thanks for your PR!

 Currently it do not compile:
```
c/enc/././fast_log.h:16:8: error: C++ style comments are not allowed in ISO C90
  #endif // defined(_MSC_VER)
         ^^^^^^^^^^^^^^^^^^^^
```

Besides that, we try to keep "portabililty" layer in `port.h` file. Currently, for encoder, there are 2 exclusions: 'fast log 2' and 'find match length'. Going to comb them in near future as well. Would you mind if I move your improvement to `port.h`?

KindDragon:
Yep, move them

KindDragon:
Closing it because of #636

PR #620, opened 2017-11-24
Add autotools build

eustas:
Also separate semantic and ABI version.


eustas:
@robryk, @badger, could you take a look, please.

bagder:
I needed this patch, as otherwise automake complains on the lack of a three files. NEWS, AUTHORS and Changelog.

~~~diff
diff --git a/Makefile.am b/Makefile.am
index 55ba9de..e823159 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,6 +1,6 @@
-AUTOMAKE_OPTIONS = gnu nostdinc subdir-objects
+AUTOMAKE_OPTIONS = foreign nostdinc subdir-objects
 
 ACLOCAL_AMFLAGS = -I m4
 
 LIBBROTLI_VERSION_INFO = -version-info @PACKAGE_VERSION@

bagder:
An additional suggestion: put the headers and source files in a separate file that can be included by both cmake and automake (if you intend to keep maintaining both): I've found that to work really well in the curl project. We put them in [Makefile.inc](https://github.com/curl/curl/blob/master/lib/Makefile.inc) files.

eustas:
Done. Thanks.

PR #623, opened 2017-11-29
Fix missing symbols errors in libbrotlienc and dec

Sp1l:
When using autotools to build the binary and libraries, the resulting libraries don't link `brotlicommon` or `m`.
This was detected when building cURL 7.57.0 which has now has brotli support. During configure it was failing
```
checking run-time libs availability... failed
configure: error: one or more libs available at link-time are not available run-time. Libs used at link-time: -lbrotlidec -lz -L/usr/local/lib
```
inspection of config.log showed missing symbols from libbrotlicommon as the cause.

This patch results in the encryption and decryption libs to be properly linked against libbrotlicommon and libm.
See also https://bugs.freebsd.org/223966


eustas:
Thanks!

PR #627, opened 2017-12-04
Work around Linuxisms

Sp1l:
Missed this in my previous tests. Sorry for that.

On BSDs, both bc and sed are part of the base operating system. For sed this results in an error as the check construct (--version) is a GNU-ism and only works for GNU sed, not for bsd sed.
Similarly, BSD sed does not take parameters after the filename(s) operated on. Moving `-i` to the front fixes that. `-r` is provided for GNU compat in BSD sed as an alias of `-E`. The `-i` option in BSD sed requires an extension to work in-place.

(thank you for picking up the nginx module too!)

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -2,7 +2,9 @@
 
 REQUIRED='is required, but not installed.'
 bc -v >/dev/null 2>&1 || { echo >&2 "'bc' $REQUIRED"; exit 1; }
+if [ `uname -s` != "FreeBSD" ]; then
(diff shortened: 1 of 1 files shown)


eustas:
Thanks!

felixbuenemann:
@Sp1l Btw. if the FreeBSD version of sed behaves like the macOS one, you can do in-place replacements without backups by using `-i ''`, however that leads to a "No such file or directory" error on GNU sed, so for compatibility with both versions a backup extension is required.

PR #629, opened 2017-12-07
Don't set rpath

jbicha:
This is based on a Debian [patch](https://anonscm.debian.org/cgit/collab-maint/brotli.git/tree/debian/patches/0001-Do-not-set-rpath.patch).

Debian doesn't like [rpath](https://lintian.debian.org/tags/binary-or-shlib-defines-rpath.html). I believe Fedora has similar concerns.

CC @fred-wang

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,16 +24,6 @@ mark_as_advanced(BROTLI_BUNDLED_MODE)
 
 include(GNUInstallDirs)
 
-# When building shared libraries it is important to set the correct rpath.
-# See https://cmake.org/Wiki/CMake_RPATH_handling#Always_full_RPATH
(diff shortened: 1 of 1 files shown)


jbicha:
I signed it!

fred-wang:
When I was trying to build brotli/woff2 and to run the executables (I don't remember exactly whether it was after a local "make install" or not), the shared libraries were not found so I had to copy this from the CMake doc. Since my goal with CMake was to make brotli/woff2 easily buildable for distro maintainers and that libraries packages will be installed in standard system paths anyway, this RPATH option should not be necessary indeed. I guess people not willing to use the libraries installed on the system will just link static versions. So this is fine by me.

eustas:
Thanks!

PR #636, opened 2018-02-08
Use MSVC intrinsics in Log2FloorNonZero and FindMatchLengthWithLimit

saucecontrol:
This PR finishes off the work started in https://github.com/google/brotli/pull/618

Switched compiler intrinsic code to use more generic switches and added MSVC-specific versions of the clz and ctz calls where the GCC __builtin functions are used.

Preliminary testing shows an almost 10% speed improvement for encoding on Windows x64.

### DIFF
--- a/c/enc/find_match_length.h
+++ b/c/enc/find_match_length.h
@@ -17,8 +17,7 @@ extern "C" {
 #endif
 
 /* Separate implementation for little-endian 64-bit targets, for speed. */
-#if defined(__GNUC__) && defined(_LP64) && defined(BROTLI_LITTLE_ENDIAN)
-
+#if defined(BROTLI_TZCNT64) && BROTLI_64_BITS && BROTLI_LITTLE_ENDIAN
 static BROTLI_INLINE size_t FindMatchLengthWithLimit(const uint8_t* s1,
                                                      const uint8_t* s2,
                                                      size_t limit) {
@@ -32,7 +31,7 @@ static BROTLI_INLINE size_t FindMatchLengthWithLimit(const uint8_t* s1,
     } else {
       uint64_t x = BROTLI_UNALIGNED_LOAD64LE(s2) ^
           BROTLI_UNALIGNED_LOAD64LE(s1 + matched);
-      size_t matching_bits = (size_t)__builtin_ctzll(x);
+      size_t matching_bits = (size_t)BROTLI_TZCNT64(x);
       matched += matching_bits >> 3;
       return matched;
     }
--- a/c/common/platform.h
+++ b/c/common/platform.h
@@ -41,6 +41,10 @@
 #define BROTLI_X_BIG_ENDIAN BIG_ENDIAN
 #endif
 
+#if BROTLI_MSVC_VERSION_CHECK(12, 0, 0)
+#include <intrin.h>
+#endif
+
 #if defined(BROTLI_ENABLE_LOG) || defined(BROTLI_DEBUG)
 #include <assert.h>
 #include <stdio.h>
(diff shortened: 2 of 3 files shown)


saucecontrol:
I signed it!

matbech:
I recommend to use _tzcnt_u64 on x86 and x64 instead of _BitScanForward64 as it improves performance on AMD.
See the explanation here: https://github.com/matbech/zlib/blob/6f8722e567a9e91c6d6662260bbae89eba5dd9a6/deflate.c#L1270

saucecontrol:
True, but that's a BMI instruction, so you have to add a cpuid check and still need the fallback to `bsf`

matbech:
No, this is not necessary. See the comments in the previous link:
"On processors that do not support TZCNT, the instruction byte encoding is executed as BSF. In this case the result for 0 is undefined."

saucecontrol:
I saw that comment, but I was looking at the VS docs, which have BMI listed as a requirement.  I tried it out on an older processor and it does indeed work as explained here: https://stackoverflow.com/questions/43880227/why-does-tzcnt-work-for-my-sandy-bridge-processor

c/enc/find_match_length.h
```diff
@@ -32,7 +34,14 @@ static BROTLI_INLINE size_t FindMatchLengthWithLimit(const uint8_t* s1,
     } else {
       uint64_t x = BROTLI_UNALIGNED_LOAD64LE(s2) ^
           BROTLI_UNALIGNED_LOAD64LE(s1 + matched);
+#if BROTLI_GNUC_HAS_BUILTIN(__builtin_ctzll, 3, 4, 0) || \
+    BROTLI_INTEL_VERSION_CHECK(16, 0, 0)
       size_t matching_bits = (size_t)__builtin_ctzll(x);
+#else /* BROTLI_MSVC_INTRINSICS */
+      uint32_t first_one;
+      _BitScanForward64(&first_one, x);
```
KindDragon: `_BitScanForward64` only supported with 64-bit compiler (`defined(_M_AMD64)`). I think you should use `_BitScanForward`

Something like that:
```
  size_t matching_bits;
  if (_BitScanForward(&first_one, x))
    matching_bits = (size_t)first_one;
  else if (_BitScanForward(&first_one, (x >> 32)))
    matching_bits = (size_t)(where + 32);
```

saucecontrol: The whole function body is wrapped in `#if BROTLI_64_BITS`, so that `#else` would never happen.  I've refactored the change and used `_tzcnt_u64`, so hopefully it's a more appealing PR now.

saucecontrol:
Travis failure appears to be a config issue.

@eustas I refactored these changes to be more in keeping with the current platform.h style.  Is there anything else I can do to make this PR more appealing?

eustas:
Looks nice! I'll try to fix travis java builds soon and then land this PR. Thanks!

Just one question - what is "c" function/macro (line 540)?

saucecontrol:
> what is "c" function/macro (line 540)?

Oh dear, that was a ctrl-c gone terribly wrong when I was triple-checking the argument types of that intrinsic.  That case only applies in the unlikely event someone uses MSVC for an ARM64 build, so I could only test it by breaking the condition checking for X64.  All fixed now.

Thanks for the review!

c/enc/find_match_length.h
KindDragon: Not all processors support `_tzcnt_u64` (`BMI1 is available in AMD's Jaguar,[4] Piledriver[5] and newer processors, and in Intel's Haswell[6] and newer processors. `)

I think better revert to _BitScanForward

saucecontrol: See the earlier comments on this PR.  TZCNT decodes as BSF on non-BMI1 processors
https://stackoverflow.com/questions/43880227/why-does-tzcnt-work-for-my-sandy-bridge-processor

eustas:
Awesome PR! Thanks!

eustas:
Hey, @saucecontrol 
Does `_tzcnt_uXX` require including ammintrin.h / immintrin.h?

eustas:
(see #837)

saucecontrol:
The documented behavior (see intro paragraph [here](https://docs.microsoft.com/en-us/cpp/intrinsics/x86-intrinsics-list?view=vs-2015#x86-intrinsics)) is that

> The intrin.h header includes both immintrin.h and ammintrin.h for simplicity.

As to #837, I'm not familiar at all with Visual C++ for Python, but I see from its [download page](https://www.microsoft.com/en-us/download/details.aspx?id=44266) that it was never an officially supported project.

> This compiler package is entirely unsupported and represents a snapshot of the last version that is fully compatible with Python 2.7. For the latest security and bug fixes, please use the latest version of Microsoft Visual Studio for any development work that does not directly interface with Python 2.7.

Judging from its download size, it may not have been a complete implementation.  And since Python 2 is now out of support itself, it doesn't seem like it's worth much investigation time.

PR #645, opened 2018-03-07
Fix soname version

mcatanzaro:
0ad94eed00420bf1154cb16a289aa27efbb30c01 changed the libbrotli libraries' soname
version from 1.0.1 (following standard conventions) to just 1. This isn't the end of the world,
but it seems like a mistake.

See also: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/O6W3HE445NYZRVVFGLA2OTAR7ZHRFHI2/

Note: this is 100% totally untested: I just skimmed though the git history looking for suspicious commits and decided to undo the suspicious thing. But I think it should be fine.


mcatanzaro:
I'm not going to sign the CLA... this change is clearly de minimis; feel free to take it and re-commit or whatever necessary to satisfy the legal overlords.

kkofler:
(Another Fedora packager here, a KDE packager who is thus more familiar with CMake. ;-) )

Your patch is clearly wrong. If anything, you would want to set both `VERSION` and `SOVERSION` to the full `${BROTLI_ABI_COMPATIBILITY}.${BROTLI_ABI_AGE}.${BROTLI_ABI_REVISION}`. (It does not make sense to have `VERSION` shorter than `SOVERSION`.) But if you actually read the commit message of commit 0ad94eed00420bf1154cb16a289aa27efbb30c01, in particular: *separate semantic and ABI version*, and if you look at the name of the macro `${BROTLI_ABI_COMPATIBILITY}`, it should be pretty clear that having the `SOVERSION` set as just `${BROTLI_ABI_COMPATIBILITY}` is intended, even if the jump from `1.0.1` to just `1` may look weird at first glance. So IMHO this pull request can be discarded completely.

mcatanzaro:
> Your patch is clearly wrong. If anything, you would want to set both VERSION and SOVERSION to the full ${BROTLI_ABI_COMPATIBILITY}.${BROTLI_ABI_AGE}.${BROTLI_ABI_REVISION}. (It does not make sense to have VERSION shorter than SOVERSION.)

I will trust you on this, because I don't know what VERSION does, but I'd like to understand. I looked at [the documentation](https://cmake.org/cmake/help/v3.10/prop_tgt/VERSION.html) and didn't see any indication of what VERSION actually does. What does it do? Why should it never be shorter than SOVERSION? Why are they related at all?

> But if you actually read the commit message of commit 0ad94ee, in particular: separate semantic and ABI version, and if you look at the name of the macro ${BROTLI_ABI_COMPATIBILITY}, it should be pretty clear that having the SOVERSION set as just ${BROTLI_ABI_COMPATIBILITY} is intended, even if the jump from 1.0.1 to just 1 may look weird at first glance. So IMHO this pull request can be discarded completely.

Kevin, come on, I can read... the variables here are named CURRENT, REVISION, and AGE. There is vanishingly small chance that these were intended to be used for any purpose other than constructing a traditional x.y.z soversion... and the commit is careful to add corresponding libtool versioning to the autotools build.

mcatanzaro:
Let's put this another way: the autotools build is obviously installing libraries with a three-component soversion based on current, revision, and age. How would you change the CMake build to do the same (like it was before)?

AdamWill:
:popcorn:

fred-wang:
> I'm not going to sign the CLA... this change is clearly de minimis; feel free to take it and re-commit or whatever necessary to satisfy the legal overlords.

Igalia already signed the CLA, so that should not be needed anyway.

kkofler:
The `SOVERSION` is only the major version (the equivalent of `LT_CURRENT`). It is called this way because this is the only part actually encoded in `DT_SONAME`. This is the version that you should bump if and only if the ABI changes, as a bump forces everything dependent on the library to be rebuilt. In libtool, this must be a single integer. CMake allows you to set `SOVERSION` to anything, but it is typically an integer as in libtool.

The `VERSION` is the full version, with which the real file (as opposed to a symlink) is installed. In the libtool scheme, it is of the *major.minor.revision* form, where the minor version is actually calculated through a subtraction in libtool. CMake actually allows you to set `VERSION` to anything. Here, a classic *major.minor.revision* is used.

mcatanzaro:
@AdamWill, Kevin is right. I built the code and it does indeed install the .so.x.y.z objects exactly as it should. Here is lib64 in my install tree:

```
$ ls
libbrotlicommon.so        libbrotlidec.so.1      libbrotlienc.so.1.0.3
libbrotlicommon.so.1      libbrotlidec.so.1.0.3  libbrotlienc-static.a
libbrotlicommon.so.1.0.3  libbrotlidec-static.a  pkgconfig
libbrotlicommon-static.a  libbrotlienc.so
libbrotlidec.so           libbrotlienc.so.1
```

This PR is invalid. Are you sure .1.0.3 so is not being installed in the Fedora package? If so, something weird must be going on.

mcatanzaro:
> Are you sure .1.0.3 so is not being installed in the Fedora package?

I built the old version, 1.0.1, and found in my install tree:

```
$ ls
libbrotlicommon.so        libbrotlidec.so.1.0.1  libbrotlienc-static.a
libbrotlicommon.so.1.0.1  libbrotlidec-static.a  pkgconfig
libbrotlicommon-static.a  libbrotlienc.so
libbrotlidec.so           libbrotlienc.so.1.0.1
```

So the change between 1.0.1 and 1.0.3 was clearly a progression. It was the older version that was weird.

AdamWill:
@mcatanzaro I never said the copy of the library with the full version as its filename wasn't *present*. I said the internal soname of the library changed; in the 1.0.1 package build the soname of the libraries was e.g. `libbrotlicommon.so.1.0.1`, in the 1.0.3 package it was `libbrotlicommon.so.1`. You can verify this for yourself easily enough by just downloading the packages and examining the files. Here, look:

    [adamw@adam ~]$ objdump -p /usr/lib64/libbrotlicommon.so.1.0.1  | grep SONAME
      SONAME               libbrotlicommon.so.1.0.1
    [adamw@adam ~]$ objdump -p /tmp/libbrotlicommon.so.1.0.3 | grep SONAME
      SONAME               libbrotlicommon.so.1
    [adamw@adam ~]$

(that's comparing my installed copy of the library, from brotli-1.0.1-3.fc28.x86_64 , to the copy from brotli-1.0.3-1.fc29 , which I extracted into /tmp).

mcatanzaro:
Yes.

It's good now, and we should not change anything.

AdamWill:
Right. This was a 'problem' from Fedora's perspective, in that there was an soname change that was not announced. But the change wasn't inherently wrong, and doesn't need fixing.

PR #647, opened 2018-03-08
Use __has_declspec_attribute to determine support

donny-dont:
Clang has support for `__declspec` attributes and provides a mechanism to query whether it supports them through `__has_declspec_attribute`. 

https://clang.llvm.org/docs/LanguageExtensions.html#has-declspec-attribute

### DIFF
--- a/c/include/brotli/port.h
+++ b/c/include/brotli/port.h
@@ -27,6 +27,14 @@
 #define __has_feature(x) 0
 #endif
 
+#ifndef __has_declspec_attribute
+#if defined(_WIN32)
+#define __has_declspec_attribute(x) 1
+#else
+#define __has_declspec_attribute(x) 0
+#endif
+#endif
+
 #if defined(__GNUC__) && defined(__GNUC_MINOR__)
 #define BROTLI_GCC_VERSION (__GNUC__ * 100 + __GNUC_MINOR__)
 #else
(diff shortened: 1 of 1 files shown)


c/include/brotli/port.h
```diff
@@ -27,6 +27,14 @@
 #define __has_feature(x) 0
 #endif
 
+#ifndef __has_declspec_attribute
```
eustas: As I see, it is usually applied like this:
```
#ifndef __has_declspec_attribute
#define __has_declspec_attribute(__x) 0
#endif
```

Why `_WIN32` is special?

c/include/brotli/port.h
```diff
@@ -47,7 +55,7 @@
 #define BROTLI_MODERN_COMPILER 0
 #endif
 
-#if defined(BROTLI_SHARED_COMPILATION) && defined(_WIN32)
+#if defined(BROTLI_SHARED_COMPILATION) && (__has_declspec_attribute(dllexport) && __has_declspec_attribute(dllimport))
```
eustas: Also, for more uniform code, it is better to use semantic derivative of check - i.e. define
BROTLI_DLLIMPORT and BROTLI_DLLEXPORT (like here http://llvm.org/viewvc/llvm-project/cfe/trunk/docs/LanguageExtensions.rst?view=diff&r1=223466&r2=223467&pathrev=223467) and use them in place of __declspec(xxx)

eustas:
Is there a configuration where current macros do the wrong thing, i.e.:
 * `_WIN32` defined and there is no declspec import / export?
 * `_WIN32` is **not** defined, but declspec import / export should be set?

PR #648, opened 2018-03-13
sources: Fix missing params header

wkennington:
Without the params header being listed we can't generate a working dist
tarball using autotools since it requires all of the sources be
explicitly specified.


eustas:
Please add it to `setup.py` as well and singn CLA =)
Also I add this line to both files in the next update, if you don't mind...

wkennington:
Sounds good. Yeah the addition for the next release is fine.

eustas:
Thanks again! Merged as a part of #651

PR #653, opened 2018-03-26
Tell CMake to not check for a C++ compiler

aperezdc:
By default CMake checks both for C and C++ compilers, while the latter is not needed. Setting the list of languages to just `C` in the call to `project()` removes the unneeded check.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 # support 2.8.7.
 cmake_minimum_required(VERSION 2.8.6)
 
-project(brotli)
+project(brotli C)
 
 # If Brotli is being bundled in another project, we don't want to
 # install anything.  However, we want to let people override this, so


aperezdc:
JFTR, I found out that the unneeded C++ check was being done thanks to [this report](http://autobuild.buildroot.net/results/4e661d663b487bbbcc08a63e136d991931a28639/build-end.log) from Buildroot's autobuilder :wrench:

fred-wang:
Thanks, this LGTM. @eustas can you please review this?

Cmake reference: https://cmake.org/cmake/help/v3.11/command/project.html

eustas:
I believe CMake 2.8.6 do not recognize `LANGUAGES` keyword. We can omit it, because there is no `VERSION` part.

aperezdc:
@eustas If you need to support CMake 2.8, [according to the documentation](https://cmake.org/cmake/help/v2.8.12/cmake.html#command:project) it is possible to use:

```cmake
project(brotli C)
```

I'll check whether omitting the `LANGUAGES` keyword works also in more recent versions (my local version is 3.10.x) and update the PR.

aperezdc:
The plain `project(brotli C)` worked fine for me with CMake 3.10.3, so I have updated the PR. Hopefully the CI will pass as well :wink:

eustas:
Thanks!

Issue #542, opened 2017-04-26
--disable-shared-libs has no effect


JoeUser78:
Compiling 0.6.0 with cmake and --disable-sharesd-libs as documented in the README has absolutly no effect.
What is the correct way to build static libs and binary for 0.6.0 with cmake?

JoeUser78:
Corresponding FreeBSD bugreport:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=218851

eustas:
Going to investigate ASAP (== tomorrow)

eustas:
Current workaround for simple build cases would be:
 * `cmake . && make` for static libraries
 * `cmake . -DBUILD_SHARED_LIBS=ON && make` for shared libraries

Digging further, why this machinery doesn't work as expected.

JoeUser78:
OK, -DBUILD_SHARED_LIBS=OFF -DBUILD_STATIC_LIBS=ON did it, thanks.

eustas:
Thanks for the report. Going to update readme, and, most probably, release 0.6.1 after fixing buildfiles.

Line `CMAKE_DEPENDENT_OPTION(BUILD_SHARED_LIBS "Build shared libraries" ON "NOT BROTLI_BUNDLED_MODE" OFF)` is read as "if not in bundled mode (== top level project) then, by default, set BUILD_SHARED_LIBS=ON". For some reason, this magic didn't happen. That is what configure-cmake flag has no effect.

JoeUser78:
Thanks for your realy fast response and fix.

eustas:
Oh, I have a guess what has happened! CMake is a very nasty build system. If you occasionally use it in root project folder - you poison it. CMake will cache some variables and do unexpected things based on cached values.

That is why there is `mkdir out && cd out` and `mkdir out-static && cd static-out` in readme.

To cure the poisoned repository try `git reset --hard && git clean -dfx`.

JoeUser78:
My builds had always been out-of-source builds (FreeBSDs Ports Makefile option `USES=cmake:outsource`), so that this is not the course of the issue. Also i'm not using git but the released source code archive (`v0.6.0.tar.gz`).

osokin:
Hi,

actually we'd like to see both types of libraries (shared and static) like other ports do.

Is there any chance to see three types of a build, i.e. SHARED, STATIC, COMBINED (SHARED_AND_STATIC).

Thanks in advance.

eustas:
[...]

PR #655, opened 2018-03-26
CMake: Allow using BUILD_SHARED_LIBS to choose static/shared libs

aperezdc:
By convention projects using CMake which can build either static or shared libraries use a `BUILD_SHARED_LIBS` flag to allow selecting between both: the `add_library()` command automatically switches between both using this variable when the library kind is not passed to `add_library()`. It is also usual to expose the `BUILD_SHARED_LIBS` as an user-facing setting with the `option()` command.

This way, the following will both work as expected:

```
   % cmake -DBUILD_SHARED_LIBS=OFF ...
   % cmake -DBUILS_SHARED_LIBS=ON ...
```

This is helpful for distributions which need (or want) to build only static libraries.

----

The above being said, I understand that there might be some reason why it is desirable that both kinds of libraries are always built. If that is the case, let me know and I'll try to come up with some other way of allowing to only build static libraries, and installing them in a way that other projects trying to link with `-lbrotli*` (and/or using `pkg-config`) will find the static libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,8 @@ cmake_minimum_required(VERSION 2.8.6)
 cmake_policy(SET CMP0048 NEW)
 project(brotli C)
 
+option(BUILD_SHARED_LIBS "Build shared libraries" ON)
+
 if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   message(STATUS "Setting build type to Release as none was specified.")
   set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the type of build." FORCE)
@@ -136,10 +138,6 @@ set(BROTLI_LIBRARIES_CORE brotlienc brotlidec brotlicommon)
 set(BROTLI_LIBRARIES ${BROTLI_LIBRARIES_CORE} ${LIBM_LIBRARY})
 mark_as_advanced(BROTLI_LIBRARIES)
 
-set(BROTLI_LIBRARIES_CORE_STATIC brotlienc-static brotlidec-static brotlicommon-static)
-set(BROTLI_LIBRARIES_STATIC ${BROTLI_LIBRARIES_CORE_STATIC} ${LIBM_LIBRARY})
-mark_as_advanced(BROTLI_LIBRARIES_STATIC)
-
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
   add_definitions(-DOS_LINUX)
 elseif(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
@@ -162,29 +160,25 @@ transform_sources_list("scripts/sources.lst" "${CMAKE_CURRENT_BINARY_DIR}/source
 include("${CMAKE_CURRENT_BINARY_DIR}/sources.lst.cmake")
 
 if(BROTLI_EMSCRIPTEN)
-  set(BROTLI_SHARED_LIBS "")
-else()
-  set(BROTLI_SHARED_LIBS brotlicommon brotlidec brotlienc)
-  add_library(brotlicommon SHARED ${BROTLI_COMMON_C})
-  add_library(brotlidec SHARED ${BROTLI_DEC_C})
-  add_library(brotlienc SHARED ${BROTLI_ENC_C})
+  set(BUILD_SHARED_LIBS OFF)
 endif()
 
-set(BROTLI_STATIC_LIBS brotlicommon-static brotlidec-static brotlienc-static)
-add_library(brotlicommon-static STATIC ${BROTLI_COMMON_C})
-add_library(brotlidec-static STATIC ${BROTLI_DEC_C})
-add_library(brotlienc-static STATIC ${BROTLI_ENC_C})
+add_library(brotlicommon ${BROTLI_COMMON_C})
+add_library(brotlidec ${BROTLI_DEC_C})
+add_library(brotlienc ${BROTLI_ENC_C})
 
 # Older CMake versions does not understand INCLUDE_DIRECTORIES property.
 include_directories(${BROTLI_INCLUDE_DIRS})
 
-foreach(lib IN LISTS BROTLI_SHARED_LIBS)
-  target_compile_definitions(${lib} PUBLIC "BROTLI_SHARED_COMPILATION" )
-  string(TOUPPER "${lib}" LIB)
-  set_target_properties (${lib} PROPERTIES DEFINE_SYMBOL "${LIB}_SHARED_COMPILATION")
-endforeach()
+if(BUILD_SHARED_LIBS)
+  foreach(lib ${BROTLI_LIBRARIES_CORE})
+    target_compile_definitions(${lib} PUBLIC "BROTLI_SHARED_COMPILATION" )
+    string(TOUPPER "${lib}" LIB)
+    set_target_properties (${lib} PROPERTIES DEFINE_SYMBOL "${LIB}_SHARED_COMPILATION")
+  endforeach()
+endif()
 
-foreach(lib IN LISTS BROTLI_SHARED_LIBS BROTLI_STATIC_LIBS)
+foreach(lib ${BROTLI_LIBRARIES_CORE})
   target_link_libraries(${lib} ${LIBM_LIBRARY})
   set_property(TARGET ${lib} APPEND PROPERTY INCLUDE_DIRECTORIES ${BROTLI_INCLUDE_DIRS})
   set_target_properties(${lib} PROPERTIES
@@ -201,9 +195,6 @@ target_link_libraries(brotlidec brotlicommon)
 target_link_libraries(brotlienc brotlicommon)
 endif()
 
-target_link_libraries(brotlidec-static brotlicommon-static)
-target_link_libraries(brotlienc-static brotlicommon-static)
-
 # For projects stuck on older versions of CMake, this will set the
 # BROTLI_INCLUDE_DIRS and BROTLI_LIBRARIES variables so they still
 # have a relatively easy way to use Brotli:
@@ -217,7 +208,7 @@ endif()
 
 # Build the brotli executable
 add_executable(brotli ${BROTLI_CLI_C})
-target_link_libraries(brotli ${BROTLI_LIBRARIES_STATIC})
+target_link_libraries(brotli ${BROTLI_LIBRARIES})
 
 # Installation
 if(NOT BROTLI_BUNDLED_MODE)
@@ -234,13 +225,6 @@ if(NOT BROTLI_BUNDLED_MODE)
       RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
     )
   endif()  # BROTLI_EMSCRIPTEN
-
-  install(
-    TARGETS ${BROTLI_LIBRARIES_CORE_STATIC}
-    ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
-    LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
-    RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
-  )
 
   install(
     DIRECTORY ${BROTLI_INCLUDE_DIRS}/brotli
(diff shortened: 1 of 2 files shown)


aperezdc:
For some more context (and an example), I am trying to address building for an Linux uCLibc target which uses only static libraries, like in [this Buildroot build](http://autobuild.buildroot.net/results/f1c4b5aeb12af7b7a3e8ae01c219004ecd9befd6/) — and in general making it easier for getting only static libraries as a result of the build, with the same base names as the shared libraries.

eustas:
Can't remember the full motivation, but we moved to static+shared build in #599.
Every output artifact is covered with separate build target, so building just static libraries is easy:
`make brotlidec-static brotlienc-static`

aperezdc:
@eustas: I saw the `brotlidec-static` and `brotlienc-static` targets, but using those means that every distributor needs to manually adapt build recipes to only build those two targets. Which I can agree would be acceptable.

But then distributors need to trust that renaming libraries (`libbrotli-static.a` → `libbrotli.a`, and so on) will keep working, and that the `.pc` file will be valid to find the renamed libraries, or generate their own. IMHO it would be desirable that the output from `install` target of the build system is directly useable — both for static and shared library builds.

Do you think it would be more acceptable to leave both targets for static *and* shared libraries, then when `cmake -DBUILD_SHARED_LIBS=OFF` is used:

- Disable the targets for shared libraries.
- Static libraries are built with their current names (`lib*-static.a`).
- The `install` adds symlinks `lib*.a` → `lib*-static.a`.

Once again, thank you for your time, and let me be clear that I just want to try to make it easier for packagers to adopt the library, and that I don't have anything against building both version by default :wink:

eustas:
Perhaps the change was done to fix #542. Going to dig in further. Conceptually, there are no "contra" against this shared/static switch, while all parties (developers/distros) are happy =)

aperezdc:
In case it helps: I have been looking a bit at the [woff2](https://github.com/google/woff2) CMake build definitions, and it does use `BUILD_SHARED_LIBS` as proposed in this PR.

eustas:
Here is a [commit](https://github.com/google/brotli/commit/fb139e49cc415be9fdfe8ae7e7357e9376a47f7d) that would allow disabling shared/static libraries build (or both =). WDYT?

eustas:
I think the reason to have both shared and static is that brotli tool is build "static" (for grab-and-go scenario), but libraries are "shared", because it is what they are created for...

aperezdc:
@eustas Coming back to this topic... The solution proposed in https://github.com/google/brotli/commit/fb139e49cc415be9fdfe8ae7e7357e9376a47f7d works fine with `BUILD_SHARED_LIBS=ON` and `BUILD_STATIC_LIBS=OFF`:

- Only shared libraries are build.
- The `.pc` files are installed and using them results in linking against the shared libraried.

But using `BUILD_SHARED_LIBS=OFF` and `BUILD_STATIC_LIBS=ON` results in no `.pc` files being installed, so other packages which use them to determine how to link against the library won't build. Moreover, the names of the static libraries have a `-static` suffix, so even if they try the usual library names for linking without relying on having `.pc` files, likely they would use e.g. `-lbrotlidec` (and not `-lbrotlidec-static`).

I think the approach in https://github.com/google/brotli/commit/fb139e49cc415be9fdfe8ae7e7357e9376a47f7d would do if the `.pc` files were also installed for the static libraries, so `pkg-config --static …` can work.

xhochy:
I would be interested in reactivating this PR. @eustas Would you be open to this approach or prefer to use your commit as stated in https://github.com/google/brotli/commit/fb139e49cc415be9fdfe8ae7e7357e9376a47f7d ? 

My motivation is that in conda-forge, we as a distribution prefer dynamic linkage as this is manageable by our toolchain and static libraries aren't used but consume unnecessary space in the packages.

aperezdc:
I have updated the patch to work with the current state of the tree, which also works for version 1.0.9—something like this is still desired for packaging in Buildroot so for now it will continue  carrying this patch to be applied after unpacking the sources.

Please let me know if there is some other way in which I could improve these changes to have them merged, I would be more than happy to iterate over the patch and have it included in Brotli :smile_cat:

aperezdc:
The CIFuzz failure is caused by the [projects/brotli/build.sh](https://github.com/google/oss-fuzz/blob/3049c50d48a8e712889104db25113c2b0be00301/projects/brotli/build.sh) script in the `oss-fuzz` repository hardcoding the `brotlidec-static`; it would need to be changed to pass `-DBUILD_SHARED_LIBS=OFF` to CMake.

delroth:
I see #766 was merged and released in 1.0.8. Is this PR now obsolete?

aperezdc:
> I see #766 was merged and released in 1.0.8. Is this PR now obsolete?

@delroth It's not the same, note that #766 only allows to disable building shared libraries, but does not remove the `-static` part the the static library names, and does not make the pkg-config `.pc` files usable with the static libraries—this PR does.

ILW000:
Hello. Are the old (`libbrotli-static.a`) and new binaries (`libbrotli.a`) the same?

aperezdc:
> Hello. Are the old (`libbrotli-static.a`) and new binaries (`libbrotli.a`) the same?

Yes, the `lib*.a` files built with this patch applied are bit-by-bit equal to the old `lib*-static.a`  files:

```
% md5sum build/*.a 
2edb88652563ba78cd4f863e4d884723  build/libbrotlicommon.a
2edb88652563ba78cd4f863e4d884723  build/libbrotlicommon-static.a
d35dbe1d59060611de85fcf05693275b  build/libbrotlidec.a
d35dbe1d59060611de85fcf05693275b  build/libbrotlidec-static.a
ff8b3edac99cc8dacb2d9427d5dc5061  build/libbrotlienc.a
ff8b3edac99cc8dacb2d9427d5dc5061  build/libbrotlienc-static.a
%
```

aperezdc:
Patch rebased, and `[RFC]` tag removed from PR title—we have been using this patch for a bit over three years in Buildroot without any issue. I am still convinced that doing this in the standard way using `BUILD_SHARED_LIBS` to let CMake determine what to build, is the correct way to do things, plus this patch also solves the issue of the `.pc` files not being installed.

nh2:
Is it currently possible to build both shared and static libraries?

This would be the ideal use case for me: Being able to control their existence independently, so that you can also build both.

Some Linux distributions aim to build both flavours where possible, so that the end developer can choose between static and dynamic linking easily.

As in: https://stackoverflow.com/questions/2152077/is-it-possible-to-get-cmake-to-build-both-a-static-and-shared-library-at-the-sam

kleisauke:
Thanks! I had to apply this patch as well:
```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -221,14 +221,12 @@ if(NOT BROTLI_BUNDLED_MODE)
     RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
   )
 
-  if(NOT BROTLI_EMSCRIPTEN)
-    install(
-      TARGETS ${BROTLI_LIBRARIES_CORE}
-      ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
-      LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
-      RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
-    )
-  endif()  # BROTLI_EMSCRIPTEN
+  install(
+    TARGETS ${BROTLI_LIBRARIES_CORE}
+    ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+    LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+    RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
+  )
 
   install(
     DIRECTORY ${BROTLI_INCLUDE_DIRS}/brotli
```
To ensure that the libraries are still installed when compiling with Emscripten.

CMakeLists.txt
```diff
@@ -166,29 +164,25 @@ transform_sources_list("scripts/sources.lst" "${CMAKE_CURRENT_BINARY_DIR}/source
...
 # Older CMake versions does not understand INCLUDE_DIRECTORIES property.
 include_directories(${BROTLI_INCLUDE_DIRS})
 
-foreach(lib IN LISTS BROTLI_SHARED_LIBS)
-  target_compile_definitions(${lib} PUBLIC "BROTLI_SHARED_COMPILATION" )
-  string(TOUPPER "${lib}" LIB)
-  set_target_properties (${lib} PROPERTIES DEFINE_SYMBOL "${LIB}_SHARED_COMPILATION")
-endforeach()
+if(BUILD_SHARED_LIBS)
+  foreach(lib brotlicommon brotlidec brotlienc)
```
eustas, line 178: Please replace `brotlicommon brotlidec brotlienc` with `${BROTLI_LIBRARIES_CORE}`

CMakeLists.txt
```diff
@@ -166,29 +164,25 @@ transform_sources_list("scripts/sources.lst" "${CMAKE_CURRENT_BINARY_DIR}/source
...
+    target_compile_definitions(${lib} PUBLIC "BROTLI_SHARED_COMPILATION" )
+    string(TOUPPER "${lib}" LIB)
+    set_target_properties (${lib} PROPERTIES DEFINE_SYMBOL "${LIB}_SHARED_COMPILATION")
+  endforeach()
+endif()
 
-foreach(lib IN LISTS BROTLI_SHARED_LIBS BROTLI_STATIC_LIBS)
+foreach(lib brotlicommon brotlidec brotlienc)
```
eustas, line 185: Please replace `brotlicommon brotlidec brotlienc` with `${BROTLI_LIBRARIES_CORE}`

eustas:
LGTM modulo comments

CMakeLists.txt
aperezdc: Done, thanks!

CMakeLists.txt
aperezdc: Done as well.

aperezdc:
> Thanks! I had to apply this patch as well:
> 
> ```diff
> --- a/CMakeLists.txt
> +++ b/CMakeLists.txt
> @@ -221,14 +221,12 @@ if(NOT BROTLI_BUNDLED_MODE)
>      RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
>    )
>  
> -  if(NOT BROTLI_EMSCRIPTEN)
> -    install(
> -      TARGETS ${BROTLI_LIBRARIES_CORE}
> -      ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
> -      LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
> -      RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
> -    )
> -  endif()  # BROTLI_EMSCRIPTEN
> +  install(
> +    TARGETS ${BROTLI_LIBRARIES_CORE}
> +    ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
> +    LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
> +    RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
> +  )
>  
>    install(
>      DIRECTORY ${BROTLI_INCLUDE_DIRS}/brotli
> ```
> 
> To ensure that the libraries are still installed when compiling with Emscripten.

This PR does not change the existing logic for installation (which is: when targeting Emscripten, an install target for the libraries is _not_ configured); if you want to suggest a change in that regard, that would be a separate issue/PR instead.

aperezdc:
@eustas I have submitted an update with your suggestions, PTAL. Also the Bazel CI build is failing, but this patch only touches the CMake build&mdash;likely the Bazel build failure is caused by something completely different.

kleisauke:
> when targeting Emscripten, an install target for the libraries is _not_ configured

FWIW, commit ce222e317e36aa362e83fc50c7a6226d238e03fd fixes that and this PR would cause a regression in that regard (i.e. it causes static libraries _not_ to be installed when `BROTLI_EMSCRIPTEN` is defined).

> that would be a separate issue/PR instead.

Sure, that comment was more FYI. I'll open a separate PR for that after this lands.

eustas (approved):

eustas:
Thanks!

PR #663, opened 2018-04-20
appveyor: fix issue self-upgradig pip to v10

anthrotype:
Installing with `--user` option will leave the old pip.exe script in the $PATH, but then running the latter fails because pip 10 moved 'main' to internal modules.

The fix is to force overwriting the old pip.exe by installing globally without `--user` option.

https://github.com/pypa/pip/issues/5240#issuecomment-382989420


eustas:
Thanks!

PR #690, opened 2018-06-25
Update go_library to use standard importpath

schroederc:
Instead of using go_prefix, which is deprecated, the importpath attribute is made explicit.


eustas:
LGTM, modulo making `go/BUILD` an empty file. Let's add at least a comment there that this is a golang brotli package.

schroederc:
Done.  PTAL.

eustas:
Thanks!

PR #691, opened 2018-06-29
Set visibility for !BROTLI_SHARED_COMPILATION

kasper3:
@eustas, this was discovered while adapting to v1.0.5 in corefx.


kasper3:
We have defined `BROTLI_SHARED_COMPILATION` in make to select a different code path. However, it seems like "shared compilation" does not tell us that visibility of public API would be impacted. There could be a separate out-out knob like `BROTLI_HIDE_PUBLIC_API` for folks who want to statically compile brotli and hide all API members in order to create their own wrapper and public API.

PR #695, opened 2018-07-09
Fix missing header files

wkennington:
Our dist tarball is missing hash_rolling_inc.h and
hash_composite_inc.h, which causes subsequent autotools
builds to fail. Fix this by adding it to the sources list.


eustas:
Thanks!

PR #702, opened 2018-07-24
platform: fix unaligned 64-bit accesses on AArch32

stephenkyle-ARM:
Ensures that Aarch32 Arm builds with an Armv8 compiler do not set
BROTLI_64_BITS.

This scenario is possible with ChromeOS builds, as they may use a
toolchain with the target armv7-cros-gnueabi, but with -march=armv8.
This will set __ARM_ARCH to 8 (defining BROTLI_TARGET_ARMV8), but will
also set __ARM_32BIT_STATE and not __ARM_64BIT_STATE. Without this,
illegal 64-bit non-word-aligned reads (LDRD) may be emitted.

Also fix unaligned 64-bit reads on AArch32 - STRD was still possible to
emit.

### DIFF
--- a/c/common/platform.h
+++ b/c/common/platform.h
@@ -187,7 +187,14 @@ OR:
 
 #if (defined(__ARM_ARCH) && (__ARM_ARCH == 8)) || \
     defined(__aarch64__) || defined(__ARM64_ARCH_8__)
-#define BROTLI_TARGET_ARMV8
+#define BROTLI_TARGET_ARMV8_ANY
+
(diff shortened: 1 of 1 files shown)


stephenkyle-ARM:
Relands google/brotli/699.

stephenkyle-ARM:
The CI seems to suffer from network issues often... let me try again.

stephenkyle-ARM:
Looks good now.

eustas:
Thanks!

PR #703, opened 2018-08-15
decode: faster huffman code loading on 32-bit Arm

stephenkyle-ARM:
In both symbol preloading and decoding, force the compiler to generate a
single load instead of two when loading the bits and value fields of the
HuffmanCode struct.

### DIFF
--- a/c/common/platform.h
+++ b/c/common/platform.h
@@ -180,6 +180,12 @@ OR:
 #define BROTLI_UNUSED_FUNCTION static BROTLI_INLINE
 #endif
 
+#if BROTLI_GNUC_HAS_ATTRIBUTE(aligned, 2, 7, 0)
+#define BROTLI_ALIGNED(N) __attribute__((aligned(N)))
+#else
+#define BROTLI_ALIGNED(N)
+#endif
+
 #if (defined(__ARM_ARCH) && (__ARM_ARCH == 7)) || \
     (defined(M_ARM) && (M_ARM == 7))
 #define BROTLI_TARGET_ARMV7
@@ -343,7 +349,7 @@ static BROTLI_INLINE void BrotliUnalignedWrite64(void* p, uint64_t v) {
 /* If __attribute__(aligned) is available, use that. Otherwise, memcpy. */
 
 #if BROTLI_GNUC_HAS_ATTRIBUTE(aligned, 2, 7, 0)
-typedef  __attribute__((aligned(1))) uint64_t brotli_unaligned_uint64_t;
+typedef BROTLI_ALIGNED(1) uint64_t brotli_unaligned_uint64_t;
 
 static BROTLI_INLINE uint64_t BrotliUnalignedRead64(const void* p) {
   return (uint64_t) ((brotli_unaligned_uint64_t*) p)[0];
--- a/c/dec/huffman.h
+++ b/c/dec/huffman.h
@@ -33,6 +33,12 @@ static const uint16_t kMaxHuffmanTableSize[] = {
 
 #define BROTLI_HUFFMAN_MAX_CODE_LENGTH_CODE_LENGTH 5
 
+#if ((defined(BROTLI_TARGET_ARMV7) || defined(BROTLI_TARGET_ARMV8_32)) && \
+  BROTLI_GNUC_HAS_ATTRIBUTE(aligned, 2, 7, 0))
+#define BROTLI_HUFFMAN_CODE_FAST_LOAD
+#endif
+
+#if !defined(BROTLI_HUFFMAN_CODE_FAST_LOAD)
 /* Do not create this struct directly - use the ConstructHuffmanCode
  * constructor below! */
 typedef struct {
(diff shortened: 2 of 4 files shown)


stephenkyle-ARM:
Hi. I've been doing some profiling of brotli on 32-bit arm, and found that the following code change can reduce the number of loads we do in the hot symbol preloading/decoding code path, which is good for an average 1.5% increase in decompression performance on recent out-of-order Arm cores, and an average 5% increase in decompression performance on recent in-order Arm cores. This is testing using the test cases found in the test suite for snappy (https://github.com/google/snappy/tree/master/testdata).

I'd be interested to know if you have any other thoughts on potential avenues for optimisation of brotli on arm. I see that you've already used NEON to accelerate the CommandPostDecodeLiterals stage (the memmove16 optimisation), but otherwise there doesn't seem to be much else that I can see that is both amenable to vectorisation here, but also particularly hot.

eustas:
Let's do it in smaller steps:
 1) define BROTLI_ALIGNED(N) for compilers that know the attibute
 2) implement accessors for HuffmanCode fields and use them everywhere (preferably macro, so it will be visible, that there is some magic around)
 3) declare HuffmanCode depending on platform to be either struct, or uint32_t; add alternative accessors

stephenkyle-ARM:
Thanks for the advice, @eustas, I'll get back to you with a rewrite. Just to check for point 1, you want BROTLI_ALIGNED(N) to be a macro for __attribute__((aligned(N)), and to use it in the uint32_t definition of HuffmanCode, and also anywhere else we're using the attribute (my recent bugfix for example.)

eustas:
Correct. Thanks!

stephenkyle-ARM:
Hi. Unfortunately just using `*table & 0xFF` followed by `*table >> 16` would always produce two loads, requiring me to add some extra macros to prepare a variable for loading the value into, and loading the value - `_PREPARE_TO_ACCESS` and `_UPDATE`. The use of these macros has no effect on the x86 code, as far as I can see.

stephenkyle-ARM:
I feel like the CI is suffering from some issues? Time outs, unable to install Java, a missing standard library header on OS X, and encoder tests failing, which my patches shouldn't be affecting?

stephenkyle-ARM:
Hi @eustas, did you get a chance to look at the latest version of this PR?

eustas:
Hello. At last I've back to the brotli project. Sorry for the very long waiting time. Going to look at your PR soon. Thanks.

c/dec/huffman.h
```diff
@@ -33,11 +33,60 @@ static const uint16_t kMaxHuffmanTableSize[] = {
...
+#define BROTLI_HC_UPDATE(H)
+
+/* Getters must be given a pointer to a HuffmanCode. */
+#define BROTLI_HC_GET_BITS(H) (H->bits)
+#define BROTLI_HC_GET_VALUE(H) (H->value)
+
+/* Setters must be given a HuffmanCode. */
+#define BROTLI_HC_SET_BITS(H, B) H.bits = B
+#define BROTLI_HC_SET_VALUE(H, V) H.value = V
+#else
```
eustas: "else" also should have this comment: /* BROTLI_HUFFMAN_CODE_FAST_LOAD */
perhaps a new line before and after else would improve readability.

c/dec/huffman.h
```diff
@@ -33,11 +33,60 @@ static const uint16_t kMaxHuffmanTableSize[] = {
...
+typedef BROTLI_ALIGNED(4) uint32_t HuffmanCode;
+
+#define BROTLI_HC_PREPARE_TO_ACCESS(H) uint32_t __fastload_##H
```
eustas: Going to check if we could get rid of declaration/fetch wrappers.
If in "non-fast" mode copying struct to local variable is not slower (hopefully, so), then "prepare" would be plain HuffmanCode declaration, and "update" would be assignment...

c/common/platform.h
```diff
@@ -180,6 +180,12 @@ To apply compiler hint, enclose the branching condition into macros, like this:
 #define BROTLI_UNUSED_FUNCTION static BROTLI_INLINE
 #endif
 
+#if BROTLI_GNUC_HAS_ATTRIBUTE(aligned, 2, 7, 0)
```
eustas: This piece have been already added in your previous PR. So it is time to rebase =)

c/dec/huffman.h
stephenkyle-ARM: Hi @eustas, thanks for the feedback.

So just to check, you're suggesting we make both the setters and getters access HuffmanCode (not a pointer):

```
/* Getters and setters must be given a HuffmanCode. */        
#define BROTLI_HC_GET_BITS(H) (H.bits)                        
#define BROTLI_HC_GET_VALUE(H) (H.value)
```

(and with FAST_LOAD:)

```
/* Getters and setters must be given a HuffmanCode. */   
#define BROTLI_HC_GET_BITS(H) (H & 0xFFu)                
#define BROTLI_HC_GET_VALUE(H) (H >> 16)
```

And then we drop the PREPARE_TO_ACCESS and UPDATE macros, and just do things like:

```
static BROTLI_INLINE uint32_t DecodeSymbol(uint32_t bits,                  
                                           const HuffmanCode* table,       
                                           BrotliBitReader* br) {          
  HuffmanCode huffman_code;                                                
  table += bits & HUFFMAN_TABLE_MASK;                                      
  huffman_code = *table;                                                   
  if (BROTLI_HC_GET_BITS(huffman_code) > HUFFMAN_TABLE_BITS) {             
    uint32_t nbits = BROTLI_HC_GET_BITS(huffman_code) - HUFFMAN_TABLE_BITS;
    BrotliDropBits(br, HUFFMAN_TABLE_BITS);                                
    table += BROTLI_HC_GET_VALUE(huffman_code);                            
    table += (bits >> HUFFMAN_TABLE_BITS) & BitMask(nbits);                
    huffman_code = *table;                                                 
  }                                                                        
  BrotliDropBits(br, BROTLI_HC_GET_BITS(huffman_code));                    
  return BROTLI_HC_GET_VALUE(huffman_code);                                
}
```

(Which name would you prefer for the local variable? `huffman_code`, `prefix_code`, `_table`, etc?

stephenkyle-ARM: @eustas my rewrite of the code as described above unfortunately leads to code changes on x86, and a minor regression on the x86 chromebook I'm testing on, so I'm not sure if we can get rid of the prepare/update wrappers.

c/dec/huffman.h
stephenkyle-ARM: Done.

c/common/platform.h
stephenkyle-ARM: Done.

eustas:
Using uint32_t on x86_64 results 0.3% slowdown... That is why it wasn't on-boarded in 2015... So, yep, lets currently stick with if-def alternatives. Almost everything  is nice, but:
 * let's look for nicer macro names (prepare for access / update)
 * perhaps, replace setters with "make"; it would require more logic changes in huffman.c, but will avoid the confusion when get takes pointer, but set takes struct.

stephenkyle-ARM:
Thanks for the feedback, @eustas. I'm out of office for the next few days, I'll push an update next week. (This applies to the cross-compilation PR as well.)

c/common/platform.h
```diff
@@ -490,6 +496,11 @@ static BROTLI_INLINE brotli_reg_t BrotliRBit(brotli_reg_t input) {
 static BROTLI_INLINE void BrotliRBit(void) { /* Should break build if used. */ }
 #endif  /* BROTLI_RBIT */
 
+#if ((defined(BROTLI_TARGET_ARMV7) || defined(BROTLI_TARGET_ARMV8_32)) && \
```
eustas: Sorry for nit-picking. Lets either generalize the name/purpose, or move this piece to `huffman.h` - though it is platform-dependent, it is very narrow - only for huffman tables and only for decoder.

eustas:
Almost perfect! (and sorry for slow / thorough reviews)

stephenkyle-ARM:
No problem, @eustas! It's good to get these things right.

eustas:
Thanks.

PR #704, opened 2018-08-28
Fix auto detect of bundled mode

jibsen:
Set bundled mode to ON when parent directory is not empty.

This likely never became an issue due to the (strange) way if works in CMake; if you compare an undefined variable to the empty string the result is false. So the code to set BROTLI_BUNDLED_MODE never ran, and the uses later in the file are as boolean values, where undefined is false.


eustas:
Thanks. Going to see what's wrong with travis and merge soon.

eustas:
Hi. Please rebase the PR. I've just fixed CI.

PR #709, opened 2018-09-18
Cross compilation support

stephenkyle-ARM:


stephenkyle-ARM:
Well that's unusual - the CI actually fails to use the arm-linux-gnueabihf compiler successfully, but then doesn't detect that the job failed? Will need investigation.

stephenkyle-ARM:
I've had to drop the aarch64-linux-gnu builder, because there appears to be a bug with the qemu-aarch64 available on Ubuntu trusty.

@eustas, PTAL?

eustas:
Nice!

Unfortunately, for some reason Appveyor build is unhappy: https://ci.appveyor.com/project/szabadka/brotli/build/1.0.0%231360

stephenkyle-ARM:
Ah, looks like TOOLCHAIN (which I was using to select which toolchain to use when running make) is being set in this case, but not to the correct triplet name, so that build is failing, let me rename it...

eustas:
Thanks.

PR #712, opened 2018-09-26
Reduce hasher lookups with good matches

mraszyk:
If a match cannot be further improved, we can save a hasher lookup which may constitute a significant portion of compression time on files with long matches. For instance, on BSDiff patch files, the gain in compression time can be >50% for quality 9 (e.g., on the file `none` from [google/brotli/issues/632]).

Although our condition implies that the current match cannot be further improved, the method `FindLongestMatch` is not side-effect free (it affects the statistics of successful static dictionary lookups), and thus the change affects the compressed output in rare cases.

Our benchmarks on standard corpora (calgary, canterbury, silesia) show that the change is safe in terms of compressed size and compression time.


mraszyk:
I signed the CLA.

mraszyk:
The change does not improve the compression time on most file types (except for, e.g., BSDiff patch files), so I am closing this thread.

PR #714, opened 2018-10-05
decode: fix NEON inclusion

stephenkyle-ARM:
The macro that checks for NEON support should be __ARM_NEON, not
__ARM_NEON__. [1]

AArch64 compilers define __ARM_NEON but not __ARM_NEON__.
AArch32 compilers currently seem to define both, but could be within their
rights to drop __ARM_NEON__ in future versions.

This change moves the check into the common/platform.h file, checks for
both forms, and sets BROTLI_TARGET_NEON if NEON support is available.

[1] Section 6.5.4 of the ARM C Language Extensions.
    (At the time of writing, the latest version was Release 2.1.)

### DIFF
--- a/c/common/platform.h
+++ b/c/common/platform.h
@@ -203,6 +203,10 @@ OR:
 
 #endif  /* ARMv8 */
 
+#if defined(__ARM_NEON__) || defined(__ARM_NEON)
+#define BROTLI_TARGET_NEON
+#endif
+
 #if defined(__i386) || defined(_M_IX86)
 #define BROTLI_TARGET_X86
 #endif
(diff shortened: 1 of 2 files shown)


eustas:
Thanks!

eustas:
In this PR `BROTLI_TARGET_NEON` is used before its definition is included... So... when neon codepath is enabled, `<arm_neon.h>` included...
Going to fix very soon.

stephenkyle-ARM:
Thanks for doing that @eustas. I was curious why the CI bot I added didn't pick this up, I should have a PR that fixes this shortly...

Issue #725, opened 2018-10-27
What is a default quality of brotli command?


stokito:
Nor in `brotli --help` nor in `man brotli` is not mentioned what is a default compression quality.
It just said that quality can be in range of 0-9 but later it said that -q param can be in range 0-11.
```
$ brotli -h
Usage: brotli [OPTION]... [FILE]...
Options:
  -#                          compression level (0-9)
  -q NUM, --quality=NUM       compression level (0-11)
  -Z, --best                  use best compression level (11) (default)
```
What I see from sources  in [brotli.c](./c/tools/brotli.c) in `main` method:
```c
int main(int argc, char** argv) {
  Command command;
  Context context;
  context.quality = 11;
```
So it looks like the default quality is 11 i.e. best.
1. Could you mention this somewhere?
2. Can we set the maximum quality by `-11` param?

chylex:
The default quality is mentioned:
```
  -Z, --best                  use best compression level (11) (default)
```

I think the single dash (`-0` to `-9`) quality parameters are just for convenience, the problem is that by [convention](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) a single dash parameter can only by followed by a single character, so `-10` or `-11` would break the convention.

eustas:
@chylex Thanks for the answer.

dimaqq:
So... compression level `10` is impossible from command line, is it? 🙈

PR #728, opened 2018-10-27
State the default value for all compression level options

apan1000:
This improves UX by clearly stating what the default compression level is for all compression level options.
Fix for the first issue in #725


apan1000:
Signed CLA

eustas:
Would be confusing, if help says `compression level (0-9) (default: 11)`
Let's only update `-q` section and alter `-Z` section, so it does not mention that it is default.

PR #730, opened 2018-11-02
Ensure decompression consumes all input

jridgewell:
If not, it's a corrupt stream.

### DIFF
--- a/python/_brotli.cc
+++ b/python/_brotli.cc
@@ -414,7 +414,7 @@ static BROTLI_BOOL decompress_stream(BrotliDecoderState* dec,
       (*output).insert((*output).end(), buffer, buffer + buffer_length);
     }
   }
-  ok = result != BROTLI_DECODER_RESULT_ERROR;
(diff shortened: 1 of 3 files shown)


python/_brotli.cc
```diff
@@ -414,7 +414,7 @@ static BROTLI_BOOL decompress_stream(BrotliDecoderState* dec,
       (*output).insert((*output).end(), buffer, buffer + buffer_length);
     }
   }
-  ok = result != BROTLI_DECODER_RESULT_ERROR;
+  ok = result != BROTLI_DECODER_RESULT_ERROR && !available_in;
```
eustas: This check should be done on a different level.
This function is capable of decompressing "streams" - i.e. input could be passed in chunks.

eustas: Please ignore. Though input could be passed in chunks, there are guarantees, that while there is enough output space, decoder will consume all the brotli stream input.

PR #733, opened 2018-12-11
Improved BrotliParseAsUTF8 function

PopeSpaceous:
Hi, I was working on optimizing a compression software and was looking into brotli. 
Using some benchmarking tools I found that there was room for improvement in the BrotliParseAsUTF8 function. 
I have made some changes to the code and when you run files around a GB large you gain about a 20% improvement in speed, with no changes to the file size or changes to the file themselves. 
I have outlined more of my testing, benchmarking and code on my site if you wanted to use those as a reference for my changes:
[Benchmarking](http://compileofcrap.com/SPO600/8.html)
[Implimentation](http://compileofcrap.com/SPO600/9.html)

### DIFF
--- a/c/enc/utf8_util.c
+++ b/c/enc/utf8_util.c
@@ -24,40 +24,19 @@ static size_t BrotliParseAsUTF8(
     }
   }
   /* 2-byte UTF8 */
-  if (size > 1u &&
-      (input[0] & 0xE0) == 0xC0 &&
-      (input[1] & 0xC0) == 0x80) {
-    *symbol = (((input[0] & 0x1F) << 6) |
-               (input[1] & 0x3F));
-    if (*symbol > 0x7F) {
+  if (size > 1u) {
+    *symbol = 0x80; 
       return 2;
-    }
   }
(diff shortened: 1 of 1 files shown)


eustas:
This change is equivalent to:
```
BROTLI_BOOL BrotliIsMostlyUTF8(
     const uint8_t* data, const size_t pos, const size_t mask,	     const uint8_t* data, const size_t pos, const size_t mask,
     const size_t length, const double min_fraction) {	     const size_t length, const double min_fraction) {
  return BROTLI_TRUE;
}
```

and completely removing BrotliParseAsUTF8. And it will be even faster... But incorrect. This method is used to check if the compressed data is "mostly utf-8 data". So if your text data is textual -> result will be the same. But if it is something different, e.g. binary... compression density might suffer.

eustas:
In benchmark smaple file is 44x same piece of text. Of course `BrotliIsMostlyUTF8` will show up on top: compressor does not do anything useful, it just copies text over and over in gigantic chunks.

PR #748, opened 2019-04-09
Fix config_setting for msvc

meteorcloudy:
Route "compiler" value of config_setting through @bazel_tools/tools/cpp:compiler.
This is necessary for modern Bazel version.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -5,6 +5,8 @@ package(
     default_visibility = ["//visibility:public"],
 )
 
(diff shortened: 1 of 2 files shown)


meteorcloudy:
@eustas Can you review this? If possible, can you have a new release with this fix? This repo is used in Angular.

meteorcloudy:
Oh, I noticed this is fixed from internally, can you do a push to github?

eustas:
Will be fixed with #749

PR #752, opened 2019-04-30
Add sudo

snyff:
Since building the binary should be done using a low-privileges account, administrator will need to sudo (or any other way to escalate privileges) to install brotli.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ The basic commands to build, test and install brotli are:
     $ ../configure-cmake
     $ make
     $ make test
-    $ make install
+    $ sudo make install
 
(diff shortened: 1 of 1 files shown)


snyff:
I signed it!

eustas:
Thanks for your contribution.

Unfortunately, it looks wrong to advice users to use 'sudo'. Some might get suspicious.
Usually, non-priviliged users already [know](https://askubuntu.com/questions/2167/how-to-build-application-without-sudo-privileges) that they have to set prefix to their user dir. Privileged users have "sudo" reflex for "make install" =)

PR #737, opened 2018-12-27
Remove unnecessary var init

daisuke-t-jp:

PR #753, opened 2019-05-02
Update

eustas:
* fix executable mode of decode.js
 * explain clang-analyser about non-nullability
 * fix "dead assignment" (see also #737)
 * rename proguard.cfg -> proguard.pgcfg


eustas:
Will fix  go-bazel build later

PR #767, opened 2019-08-12
Add oss-fuzz fuzzing status badge to README

ammaraskar:
We just added fuzzing status badges in oss-fuzz.

This should provide quick at-a-glance information to make sure that the oss-fuzz build is still successful and the project is being continuously fuzzed.


eustas:
Thanks

PR #770, opened 2019-08-18
Remove the "Only Python 2.7+" line from README

mvasilkov:
Hi!

The README states that "Only Python 2.7+ is supported", while there are Python3 packages [hosted on PyPI](https://pypi.org/project/Brotli/#files) and elsewhere (for example, [Debian repos](https://packages.debian.org/stretch/python3-brotli) also have the python3-brotli package).

So I think this line is outdated, let's drop it.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -1,8 +1,6 @@
 This directory contains the code for the Python `brotli` module,
 `bro.py` tool, and roundtrip tests.
 
-Only Python 2.7+ is supported.
-
 We provide a `Makefile` to simplify common development commands.
 
 ### Installation


mvasilkov:
@googlebot I signed it!

eustas:
2.7+ means the minimal required Python version is 2.7.
Thus, versions 2.7, 2.7.[1..16], 3.0,... 3.7.4 meet this requirement, 2.6.9 does not.
If we remove this requirement, someone might wonder, why the wrapper does not compile under Ubuntu 12.04...

mvasilkov:
Ah, OK, I misread this as compatible with 2.x (as opposed to 3.x). Fair enough.

PR #776, opened 2019-09-27
Add vcpkg installation instructions

grdowns:
Brotli is available as a port in vcpkg, a C++ library manager that simplifies installation for Brotli and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build Brotli, ready to include in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/brotli/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,18 @@ https://groups.google.com/forum/#!forum/brotli
 
 ### Build instructions
 
+#### Vcpkg
+
+You can download and install brotli using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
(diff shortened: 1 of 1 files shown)


eustas:
Wow! Thanks.

Could we improve CMake file to make port patch-less?

PR #780, opened 2019-11-14
Add missing "const" to a couple of kConstants

agrieve:
These showed up in a Chromium audit:
https://bugs.chromium.org/p/chromium/issues/detail?id=747064#c8

Although already effectively const, adding "const" causes the symbols to
be moved into the read-only section of the binary.

### DIFF
--- a/c/common/dictionary.c
+++ b/c/common/dictionary.c
@@ -5,12 +5,13 @@
 */
 
 #include "./dictionary.h"
+#include "./platform.h"
 
 #if defined(__cplusplus) || defined(c_plusplus)
 extern "C" {
 #endif
 
-#ifndef BROTLI_EXTERNAL_DICTIONARY_DATA
+#if !defined(BROTLI_EXTERNAL_DICTIONARY_DATA)
 static const uint8_t kBrotliDictionaryData[] =
 {
 116,105,109,101,100,111,119,110,108,105,102,101,108,101,102,116,98,97,99,107,99,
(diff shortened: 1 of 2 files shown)


agrieve:
@eustas - Maybe you could review?

eustas:
`BrotliSetDictionaryData` does not compile (unused parameter).
What is the benefit of making those "variables" const?

agrieve:
Whoops! Sorry about that! For some reason I thought the failure was from a timeout, but not sure why I thought that (fixed).

The benefit of const is that the symbol is put into .rodata section of the ELF rather than .data. For Chrome, this means the memory for it is file-backed and shared between processes rather than "private dirty" memory. It's a small amount of savings, but also easy to fix. It came up in an audit based on variable naming convention.

eustas:
LGTM. Will land tonight

agrieve:
Awesome! Thanks. (hopefully actually fixed compile this time)

agrieve:
🏓

c/common/dictionary.c
```diff
@@ -5895,9 +5899,13 @@ const BrotliDictionary* BrotliGetDictionary() {
 }
 
 void BrotliSetDictionaryData(const uint8_t* data) {
+#if defined(BROTLI_EXTERNAL_DICTIONARY_DATA)
   if (!!data && !kBrotliDictionary.data) {
     kBrotliDictionary.data = data;
   }
+#else
+  (void)data;  // Appease -Werror=unused-parameter
```
eustas: Use BROTLI_UNUSED macro

eustas:
Thanks.

PR #783, opened 2019-12-05
Add HAVE_LOG2 build macro

### LINKED ISSUE #781: cmake configure fails with Android NDK for api < 18 (log2() not found)
Related: #341, #351.

The underlying issue of log2 availability is fixed. However the cmake script still emits a FATAL_ERROR if the log2 function is not found at configure time:

https://github.com/google/brotli/blob/c435f066751ef83aa4194445085a70ad9d193704/CMakeLists.txt#L114

Simply removing the line is enough to build with `__ANDROID__` == 15.

axnsan12:
Fixes #781


eustas:
LGTM. Please move to comment ("Visual Studio...") to the place where this fact is detected and HAVE_LOG2 is defined.

Also, it would be nice to change HAVE_LOG2 to BROTLI_HAVE_LOG2 to avoid unprefixed defines.

Thank you, and sorry for the super long response.

eustas:
Thanks!

PR #805, opened 2020-05-12
.gitignore: Ignore .obj files

OZ1:
EDK II windows build produces .obj files in source tree


OZ1:
@googlebot I signed it!

PR #806, opened 2020-05-12
Add file opener `brotli.open` for Python

youknowone:
The goal of this PR is adding `brotli.open` like `gzip.open`.

The patch is not completed yet - but the basic function compress/decompress is done as proof of concept. `file_test.py` is testing it.

Before revising more, I want to know if this project is interested in this patch or not.


youknowone:
@googlebot I signed it!

eustas:
Looks interesting.
Unfortunately, can't find the mentioned "Andrew Kuchling's minigzip.py". Could you share a link to it?

youknowone:
I copied it from CPython stdlib. So actually I didn't see that file. This is the CPython file: `https://github.com/python/cpython/blob/master/Lib/gzip.py#L6`

youknowone:
I splited CPython contributed part as a seperated commit but it causes CLA warning. what should i do in this case?

youknowone:
By needs, I released this one as a seperated project
- https://github.com/youknowone/brotli-file
- https://pypi.org/project/brotli-file/

Because dependencies are using Python3 for now, it needs a few tasks to finish this PR.
Let me know which things are requried below. Then I can work more for the patch

- [ ] test for file interface which can be written from CPython test_gzip.py
- [ ] CPython3-only availability check for the file interface - if this is ok approach
- [ ] Python2 and other pythons comaptibility work which including CPython `_compression` port

python/brotli_file.py
```diff
@@ -0,0 +1,322 @@
...
+            raise ValueError("Invalid mode: %r" % (mode,))
+    else:
+        if encoding is not None:
+            raise ValueError("Argument 'encoding' not supported in binary mode")
+        if errors is not None:
+            raise ValueError("Argument 'errors' not supported in binary mode")
+        if newline is not None:
+            raise ValueError("Argument 'newline' not supported in binary mode")
+
+    gz_mode = mode.replace("t", "")
```
eustas, line 50: `br_mode`?

brechtm:
@youknowone This would come in very handy, so I wanted to kindly ask whether you're interested in bringing this PR to completion. Thanks!

youknowone:
@brechtm Since nobody having project ownership is interested in, I feel like spending more effort for this patch is not reasonable enough.

You can install a separated project using pip https://pypi.org/project/brotli-file/

brechtm:
Thanks, @youknowone. I tried brotli-file, but unfortunately decompression is quite a bit slower than command-line brotli. I was planning on using brotli for compressing my CSV files which I read into pandas, but it seems `pandas.read_csv()` supports zstd directly (through python-zstandard), which does not suffer from a performance hit like brotli-file. Since it has better support for Python at this point, I'll be using zstd instead.

jyrkialakuijala:
> @brechtm Since nobody having project ownership 

There is an unreplied review finding from eustas@ (with project ownership) here: https://github.com/google/brotli/pull/806/files/3462440035eb969771455d8812e5e7ed95e66a5f#r1058955829

PR #813, opened 2020-06-30
Update platform.h

fisherwky:
fix compile error (platform.h:362: error: cast discards qualifiers from pointer target type)


fisherwky:
@googlebot I signed it!

eustas:
Hello. Thanks for the PR.

I wonder, what compiler + flags combination is used? We could try to add it to CI matrix to avoid such errors in the future...

fisherwky:
Hi eustas,
There is no problem to build brotli with command "make lib" in i386 platform and x64 platform.
When I select some files to build with my own project, include brotli/c/common/*,  brotli/c/dec/*, brotli/c/include/* and found this compiler error in i386 platform (not happened in x64 platform).
I think it has something with the macro BROTLI_GNUC_HAS_ATTRIBUTE().  I hope this would be some of help.

eustas:
Thanks.

PR #823, opened 2020-08-07
Recompression

riknel:
Faster and more efficient recompression using previous compression artifacts such as backward references and block splits.


riknel:
@googlebot I fixed it.

PR #826, opened 2020-08-26
Update

eustas:
* IMPORTANT: decoder: fix overflow when input chunk is >2GiB
 * simplify max Huffman table size calculation
 * eliminate symbol duplicates (static arrays in .h files)
 * minor combing in research/ code


eustas:
@googlebot I consent.

rcsanchez97:
@eustas I am preparing an update for brotli in Debian (based on an older release, 0.5.2) to address CVE-2020-8927.  Since the PR you merged included four distinct and separate changes in a single commit it is difficult for me to determine the minimum change needed to address CVE-2020-8927.  My belief is that the specific change that I need to isolate is associated with the changelog entry `IMPORTANT: decoder: fix overflow when input chunk is >2GiB.

Though I attempted to isolate the part of the commit that was necessary I was unsure and so I consulted with another Debian developer.  His assessment of my proposed patch was that it was perhaps too broad.  Could you identify the specific hunks that are needed to address CVE-2020-8927?  Also, is there any available information regarding a proof of concept of the vulnerability or a reproducer?

deymo:
Hi @rcsanchez97 , first of all sorry I didn't see the comment before. Due to the way we have set up the import/export changes like this one it didn't generate a small commit with just the CVE fix, sorry about that.

I discussed this with my colleagues here and we think the only change on top of 1.0.7 needed to fix that CVE-2020-8927 is the one attached here. I didn't verify if there were other changes that could affect this from 0.5.2 which is quite old, but given that the fix is a cap in the bit reader's BrotliGetRemainingBytes() function I don't expect other changes to affect it.

Let me know if you need any help and thanks for your maintainer work on brotli.

[cve-2020-8927.patch.txt](https://github.com/google/brotli/files/5622909/cve-2020-8927.patch.txt)

rcsanchez97:
@deymo No worries on the delay; the issue has been around for a while and as you pointed out I am dealing with a quite old version.  Thanks for verifying the necessary components of the fix.  I will get this updated in Debian.  Your assistance is very much appreciated.

PR #834, opened 2020-08-29
add execution time in verbose mode

ahorek:
before
```
Compressed [file.xxx]: 3.114 MiB -> 3.007 MiB
```

after
```
Compressed [file.xxx]: 3.114 MiB -> 3.007 MiB in 1.98 sec
```

### DIFF
--- a/c/tools/brotli.c
+++ b/c/tools/brotli.c
@@ -132,6 +132,8 @@ typedef struct {
      until 4GiB+ files are compressed / decompressed on 32-bit CPUs. */
   size_t total_in;
   size_t total_out;
+  clock_t start_time;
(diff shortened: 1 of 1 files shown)


ahorek:
@googlebot I fixed it.

eustas:
Thanks for the PR.
Console output would be very slow... So "end time" should be taken right after the last flush (lines 900 and 968).

ahorek:
> 
> 
> Thanks for the PR.
> Console output would be very slow... So "end time" should be taken right after the last flush (lines 900 and 968).

thanks, done

eustas:
Great! Landing.

PR #740, opened 2019-02-14
Add runtime linker path to pkg-config files

nigoroll:
Otherwise libraries will not be found at runtime when installing to a path not included in the default runtime linker's path with programs linking brotli configured via pkg-config.

PR #838, opened 2020-09-02
Revert "Add runtime linker path to pkg-config files (#740)"

eustas:
This reverts commit 31754d4ffce14153b5c2addf7a11019ec23f51c1.


muelli:
Cool, thanks for the reversal. It'd be nice to have a new minor release s.t. we don't have to carry the patch.

PR #839, opened 2020-09-04
Fix clang-10 compilation issue

rojkov:
clang-10 throws the following error:
```
In file included from external/org_brotli/c/enc/bit_cost.c:9:
external/org_brotli/c/enc/./bit_cost.h:48:16: error: implicit conversion from 'size_t' (aka 'unsigned long') to 'double' may lose precision [-Werror,-Wimplicit-int-float-conversion]
  if (retval < sum) {
             ~ ^~~
1 error generated.
```
Make the conversion explicit.


eustas:
Thanks

PR #841, opened 2020-09-12
Change MANIFEST.in to include python regression tests in tarball

kmosiejczuk:
For OpenBSD, we like having the Python regression tests included in the tarball so if any dependencies change, we can easily verify that we haven't broken things. 

This change just adds those test files to the PyPI tarball.


kmosiejczuk:
@googlebot I signed it.

Issue #837, opened 2020-09-01
Brotli 1.0.9 with Python 2.7 Windows seems broken


gpotter2:
Hi, since 1.0.8 (we're actually using 1.0.9) we're getting some installation failures on automated builds that use both Python 2.7 and Windows (exclusively this case).
See below for the error log

<details>
<summary>The error log</summary>

```
Collecting brotli
  Downloading Brotli-1.0.9.zip (510 kB)
  Building wheel for brotli (setup.py): started
  Building wheel for brotli (setup.py): finished with status 'error'
  ERROR: Command errored out with exit status 1:
   command: 'C:\projects\scapy\.tox\py27-windows\Scripts\python.EXE' -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'c:\\users\\appveyor\\appdata\\local\\temp\\1\\pip-install-6b28ed\\brotli\\setup.py'"'"'; __file__='"'"'c:\\users\\appveyor\\appdata\\local\\temp\\1\\pip-install-6b28ed\\brotli\\setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d 'c:\users\appveyor\appdata\local\temp\1\pip-wheel-cv3vb2'
       cwd: c:\users\appveyor\appdata\local\temp\1\pip-install-6b28ed\brotli\
  Complete output (93 lines):
  running bdist_wheel
  running build
  running build_py
  creating bin
  creating bin\lib.win-amd64-2.7
  copying python\brotli.py -> bin\lib.win-amd64-2.7
  running build_ext
  building '_brotli' extension
  creating bin\temp.win-amd64-2.7
  creating bin\temp.win-amd64-2.7\Release
  creating bin\temp.win-amd64-2.7\Release\c
  creating bin\temp.win-amd64-2.7\Release\c\common
  creating bin\temp.win-amd64-2.7\Release\c\dec
  creating bin\temp.win-amd64-2.7\Release\c\enc
[...]

PR #843, opened 2020-09-16
Fix VC C++ 12.0 BROTLI_MSVC_VERSION_CHECK calls

gpotter2:
fix https://github.com/google/brotli/issues/837

I was able to try it out locally on Python 2.7 with VC C++ 9.0. This did fix the issue. If possible, it would be nice to have a new version available on PyPy as soon as possible, since 1.0.7 suffers from CVEs.

I really wonder why this wasn't triggered by your test base.
```
Environment: BUILD_SYSTEM=Python, PYTHON=C:\Python27, PYTHON_VERSION=2.7.x, PYTHON_ARCH=32
```
fits the description :/

### DIFF
--- a/c/common/platform.h
+++ b/c/common/platform.h
@@ -40,7 +40,7 @@
 #define BROTLI_X_BIG_ENDIAN BIG_ENDIAN
 #endif
 
-#if BROTLI_MSVC_VERSION_CHECK(12, 0, 0)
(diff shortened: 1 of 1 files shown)


gpotter2:
@googlebot I signed it!

eustas:
Oh, it seems that MS VC version was confused with _MSC_VER.
Thanks for checking and fixing!

PR #852, opened 2020-10-09
Update

eustas:
* comments and clarifications in block_splitter
 * power-of-2 aligned allocations for Hasher
 * refresh decode.js from Java sources


eustas:
@googlebot I consent.

Issue #850, opened 2020-10-02
Source tarball is missing testdata files since v1.0.8


delroth:
Something changed between v1.0.7 and v1.0.8 which caused the testdata files to disappear from the source tarball. This means that it's now impossible to run unit tests at least for Python without using a git clone instead.

delroth:
This is because all the files were marked as export-ignore except for a select few. cc @eustas who committed that change

candrews:
Here's the commit that removed the test data: https://github.com/google/brotli/commit/d0529182550129fe047a2d92a132fa14522deda2

OpenBSD requested that the test data be added back into the PyPI tarball: https://github.com/google/brotli/pull/841 and that was done: https://github.com/google/brotli/commit/f2ca32eda6508f7eeb5b50d30ac6335edec173c6 So, I think it's logical that the test data also be added back into the source tarball.

wltjr:
Much more than just the test data was removed. Other sources like Java sources were removed. They are not part of 1.0.8 or 1.0.9. I will open a new issue regarding that, and refer to this one.

olivier-fs:
Hi !

Bumping this old issue... 

1) This is still a problem with latest 1.1.0, because CMake outputs errors about missing files.

Latest CMake-3.27.4 output :
```shell
-- Building for: Visual Studio 16 2019
-- Selecting Windows SDK version 10.0.22000.0 to target Windows 10.0.19045.
-- The C compiler identification is MSVC 19.29.30151.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Build type is ''
-- Performing Test BROTLI_EMSCRIPTEN
-- Performing Test BROTLI_EMSCRIPTEN - Failed
-- Compiler is not EMSCRIPTEN
-- Looking for log2
-- Looking for log2 - found
CMake Warning at CMakeLists.txt:264 (message):
  Test file tests/testdata/alice29.txt does not exist.


CMake Warning at CMakeLists.txt:264 (message):
  Test file tests/testdata/asyoulik.txt does not exist.


CMake Warning at CMakeLists.txt:264 (message):
  Test file tests/testdata/lcet10.txt does not exist.


[...]

PR #857, opened 2020-10-18
Include testdata in tarball (#850)

candrews:
Distributions and users run tests to ensure brotli is working and testdata is required for those tests.

This commit brings the git source tarball in alignment with the PyPI tarball which already includes this testdata, see commit f2ca32e.


candrews:
@googlebot I signed it!

candrews:
See https://github.com/google/brotli/issues/850#issuecomment-1717591323

PR #858, opened 2020-10-19
fix 'include'

vorfol:
OpenVMS compiler fails if include file starts with ./


vorfol:
@googlebot I signed it!

PR #859, opened 2020-10-20
Compilation for Python on OpenVMS

vorfol:
The project is compiled on OpenVMS. At the moment the OpenVMS compiler does not allow "./" in the include path.


eustas:
After #925 most of changes ("./" -> "") from this PR will not be necessary. Let's discuss the other changes.
Namely, inttypes stub, and what are the alternatives. At least, let's keep it in one place (types.h).

PR #865, opened 2020-11-06
patch CMakeLists.txt to allow static lib

freezeboy:
Updated patch from #655 on par with master


freezeboy:
@googlebot I signed it!

eustas:
See #655

PR #874, opened 2020-12-27
Fix: Build from Bazel

sgammon:
This changeset applies fixes to enable `org_brotli` to be used directly from downstream Bazel builds.

### DIFF
--- a/java/BUILD
+++ b/java/BUILD
@@ -54,9 +54,9 @@ java_library(
 cc_binary(
     name = "brotli_jni.dll",
     srcs = [
-        "//org/brotli/wrapper/common:jni_src",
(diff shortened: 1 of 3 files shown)


eustas:
Sorry for the very late response. Please rebase.
Why do we need `repositories.bzl`

sgammon:
hey @eustas -- sorry likewise for the very late response :) `repositories.bzl` allows downstream dependencies to pull in Bazel deps and include them in their own build graph.

i don't remember why i left it empty.

what do you think of this PR in general? is this worth rebasing and fixing up as described?

sgammon:
i think we're living on a fork because of these transitive deps we need, but that doesn't end up staying in sync anyway obviously, so we'd be happy to contribute a PR

PR #877, opened 2021-01-27
unichr was removed in Python 3 because all str are Unicode

cclauss:
https://python-future.org/compatible_idioms.html#unichr


eustas:
Thanks

PR #887, opened 2021-02-15
[Java] make it possible to set modes (generic, text, font)

martin-g:


martin-g:
@googlebot I signed it!

martin-g:
Related to: https://github.com/hyperxpro/Brotli4j/pull/15

java/org/brotli/wrapper/enc/Encoder.java
```diff
@@ -22,18 +23,76 @@
...
+    TEXT(1),
+    /**
+     * Compression mode used in WOFF 2.0.
+     */
+    FONT(2);
+
+    private final int value;
+
+    Mode(int value) {
+        this.value = value;
```
eustas, line 48: Default indentation is 2 spaces in this project.

java/org/brotli/wrapper/enc/Encoder.java
```diff
@@ -22,18 +23,76 @@
...
+  /**
+   * https://www.brotli.org/encode.html#aa6f
+   * See encode.h, typedef enum BrotliEncoderMode
+   */
+  public enum Mode {
+    /**
+     * Default compression mode.
+     * In this mode compressor does not know anything in advance about the properties of the input.
+     */
+    GENERIC(0),
```
eustas, line 35: Why simply not use ordinal?

java/org/brotli/wrapper/enc/Encoder.java
```diff
@@ -22,18 +23,76 @@
...
+
+    Mode(int value) {
+        this.value = value;
+    }
+
+    public int value() {
+        return value;
+    }
+
+    public static Mode of(int value) {
```
eustas, line 55: With ordinals that becomes simply accessing values array

java/org/brotli/wrapper/enc/Encoder.java
```diff
@@ -22,18 +23,76 @@
...
+    }
+  }
+
   /**
    * Brotli encoder settings.
    */
   public static final class Parameters {
     private int quality = -1;
     private int lgwin = -1;
+    private Mode mode = Mode.GENERIC;
```
eustas, line 88: Prever to have "null" value, to represent default and does not override encoder setting in case default wasn't changed.

java/org/brotli/wrapper/enc/encoder_jni.cc
```diff
@@ -79,6 +79,10 @@ Java_org_brotli_wrapper_enc_EncoderJNI_nativeCreate(
     if (lgwin >= 0) {
       BrotliEncoderSetParameter(handle->state, BROTLI_PARAM_LGWIN, lgwin);
     }
+    int mode = context[4];
+    if (mode >= 0) {
```
eustas, line 83: `mode > 0` as with lgwin...

eustas:
Travis CI should work now.

java/org/brotli/wrapper/enc/Encoder.java
martin-g: The value should be the same as in the C code.
I think it is more clear this way. Otherwise if ordinal is used someone may shuffle/sort the modes later and break it

java/org/brotli/wrapper/enc/encoder_jni.cc
martin-g: If `mode` is null as you suggested above then what will be the value here ?

java/org/brotli/wrapper/enc/Encoder.java
martin-g: Fixed!

martin-g:
@eustas Feel free to rework the PR the way you think it would be better! As project maintainer you could even edit the branch for this PR!

java/org/brotli/wrapper/enc/encoder_jni.cc
eustas: Oh, you are right, `0` is valid value. Just like with lgwin we shoud set it to -1 if enum value is null

martin-g: OK. I will change it now!

martin-g: Done!

java/org/brotli/wrapper/enc/Encoder.java
martin-g: Done

java/org/brotli/wrapper/enc/Encoder.java
martin-g: @eustas  Should I rework it to use ordinal ?

eustas: Yes, please. I guess, the modes won't ever change. For 100% guarantees we could add a unit-test to check that enums have  ordinals that correspond to their C counterparts (hardcoded).

java/org/brotli/wrapper/enc/Encoder.java
```diff
@@ -22,18 +23,76 @@
...
+        case 1:
+          return TEXT;
+        case 2:
+          return FONT;
+        default:
+          throw new IllegalArgumentException("Unknown mode: " + value);
+      }
+    }
+
+    public static Mode of(String name) {
```
eustas, line 68: Given that default is to throw IAE, we could simply use built-in valueOf here.

eustas: (or simply drop this method)

java/org/brotli/wrapper/enc/Encoder.java
martin-g: Done

java/org/brotli/wrapper/enc/Encoder.java
martin-g: Dropped it!

martin-g:
All code review comments are addressed!
I didn't dare to touch the unit tests because I have no idea how to run them locally. I have no experience with Bazel.

eustas:
> All code review comments are addressed!
> I didn't dare to touch the unit tests because I have no idea how to run them locally. I have no experience with Bazel.

No worries about tests - I'll add them later,...

java/org/brotli/wrapper/enc/EncoderJNI.java
```diff
@@ -29,21 +29,23 @@
...
       this.context[3] = lgwin;
+      this.context[4] = mode != null ? mode.ordinal() : -1;
       this.inputBuffer = nativeCreate(this.context);
       if (this.context[0] == 0) {
         throw new IOException("failed to initialize native brotli encoder");
       }
       this.context[1] = 1;
       this.context[2] = 0;
       this.context[3] = 0;
+      this.context[4] = -1;
```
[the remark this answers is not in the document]
martin-g: Reverted this change!

eustas:
Nice. Thanks for PR and patience. Landing.

martin-g:
Thank you!

PR #889, opened 2021-02-24
fatal error: brotli/decode.h: No such file or directory

true-zero:
Fixes the `fatal error: brotli/decode.h: No such file or directory` because for some reason Golang does not respect the comment.


true-zero:
@googlebot I signed it!

micheartin:
This PR does not fix the issue. To use `cbrotli` in your Golang project, before you should compile the C implementation in your local machine. See the [README.md](https://github.com/google/brotli#build-instructions) for build instructions. See also <https://stackoverflow.com/q/47233208#53095243>.

Issue #891, opened 2021-03-08
copy nsec of mtime in CopyStat


JackJey:
when using brotli command in Makefile, it doesn't build incrementally but always re-compress full files again.
this is caused by compare time difference includes nsec of mtime, current brotli cli implementation doesn't copy nsec of mtime.

https://github.com/google/brotli/blob/master/c/tools/brotli.c#L661-L662

It's welcome to include nsec of mtime when copy file stat will solve incremental build by Make. it always compress when files are updated.

eustas:
Hmm. Let's take a look how this is done in other compressors, e.g. zlib...

pirxpilot:
If you are here looking for help with your Makefiles this seems to do the trick:

```Makefile
%.br: %
    brotli --best --force --keep --output=$@ $<
    touch --reference $< $@
```
`touch` will update the timestamp from a source file and that will stop `make` from trying to re-brotlify the same file over and over again

And yeah - some other compressors have the same problem:
```
$ stat -c "%y %n" History.md*
2022-05-01 10:13:20.795119946 +0200 History.md
2022-05-01 10:13:20.000000000 +0200 History.md.br
2022-05-01 10:13:20.000000000 +0200 History.md.bz2
2022-05-01 10:13:20.795119946 +0200 History.md.gz
2022-05-01 10:13:20.000000000 +0200 History.md.lz
```

eustas:
Will take a look soon. Thanks for the heads-up.

eustas:
Should be somewhat addressed with #992

PR #892, opened 2021-03-08
[WIP] copy nsec of atime/mtime in CopyStat #891

JackJey:
for #891 

copy tv_nsec in stat result to dest file.
switching case for APPLE & GLIBC for compatibility but I don't make sure this coves all case.
if this command need to cover windows, I think it needs more switching but I don't have windows env,
so adding [wip] for here and any comments and helps are welcome.

### DIFF
--- a/c/tools/brotli.c
+++ b/c/tools/brotli.c
@@ -663,8 +663,14 @@ static void CopyStat(const char* input_path, const char* output_path) {
 
   times[0].tv_sec = statbuf.st_atime;
   times[1].tv_sec = statbuf.st_mtime;
+
+#if defined(__GLIBC__)
   times[0].tv_nsec = statbuf.st_atim.tv_nsec;
   times[1].tv_nsec = statbuf.st_mtim.tv_nsec;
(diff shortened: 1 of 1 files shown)


ahorek:
good catch, unfortunately, this API seems to have many bugs on various platforms :-( https://github.com/kunpengcompute/gzip/blob/master/lib/utimens.c
there's also a Windows port that emulates the ```utimensat``` function via Windows-API calls, so it's possible but ugly.

JackJey:
@ahorek But if it copy's nsec correctly in some platform (not all platform), it's only a improvement for them and no breaking changes for others. So starting from partial support is reasonable for me.

ahorek:
@JackJey I would propose to
1/ move copy nsec into a separate function
2/ use conditions for known platforms that do support it
3/ and provide a fallback for platforms that don't

I'm ok with the partial support approach, but currently it breaks the build on some platforms that were supported before.

eustas:
Should be fixed by #992

PR #906, opened 2021-06-05
Enabled install when building with emscripten.

### LINKED ISSUE #886: [emscripten] There is no "install" target when build with emscripten
I found that if I build brotli using emscripten, There is no target called install.

the same problem occurs in this issues https://github.com/microsoft/vcpkg/issues/14574.

dlemstra:
This PR enables install when building with emscripten and fixes #886.


eustas:
Thank you.

PR #907, opened 2021-06-05
Create encode.h

BigGuy573:


deymo:
Creating this file in `.github/workflows/encode.h`  doesn't look correct to me.

Issue #848, opened 2020-09-25
cannot install brotli bindings on Python 2.7 anymore


jjnurminen:
Hi,

I cannot install brotli from PyPi any more, though seemingly it should still support Python 2.7. I can reproduce by creating a conda environment:

    conda create -n py27 python=2.7

Saying `pip install brotli` first tells me to install Visual C++ from http://aka.ms/vcpython27. However, after installing that and trying again, I get the following during compilation:

```
 _brotli.obj : warning LNK4197: export 'init_brotli' specified multiple times; using first specification
       Creating library bin\temp.win-amd64-2.7\Release\c/common\_brotli.lib and object bin\temp.win-amd64-2.7\Release\c/common\_brotli.exp
    encode.obj : error LNK2001: unresolved external symbol _tzcnt_u64
    static_dict.obj : error LNK2001: unresolved external symbol _tzcnt_u64
    backward_references.obj : error LNK2019: unresolved external symbol _tzcnt_u64 referenced in function FindMatchLengthWithLimit

```

I'm on Windows 10, 64-bit Python. Any hints are appreciated.

jjnurminen:
Seems to be related to the fact that the 1.0.9 Windows Python 2.7/AMD64 binary wheel is not on PyPi, so PyPi tries to build from source.

eustas:
The problem has been recently fixed. Soon I will make 1.0.10 release to fix this problem in PyPi

jjnurminen:
@eustas sounds good, thanks.

Antoine-Sc:
Hi,

I have the same problem by setting up esp-idf.
How did you fix the problem ? Which version of brotli ?

jjnurminen:
@Antoine-Sc if you install brotli 1.0.7, you will still get the binary wheel.

Antoine-Sc:
Ok. It did not work for me. I'm trying to find out why

GitMensch:
brotli 1.1 is out so this may be closed-fixed; if this issue is unexpectedly not fixed by this I'd suggest to close-not-planned as Python2 is much less relevant in 2024 than it was in 2020 and it would possibly useful to drop anything that is explicit done to support Python2.

PR #917, opened 2021-08-17
Release version 1.0.10

landfillbaby:
This is mostly needed due to #838, without which static linking is broken.
As Brotli is now part of JPEG XL, this is crucial for programs wanting to support the new file format.
The current workaround there is to use a vendored copy of brotli's master branch, which isn't really a good idea for a major standard like that.
Also note: https://github.com/google/brotli/issues/848#issuecomment-699607362

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # SECURITY NOTE
 
-Please consider updating brotli to version 1.0.9 (latest).
+Please consider updating brotli to version 1.0.9 or later.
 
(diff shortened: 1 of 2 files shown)


landfillbaby:
note: the travis failures are due to gcc-4.9 disappearing rather than any brotli problem

andrerom:
@eustas Any update on a 1.10 release?
                Or a 1.0.9.1 release without new features like shared dictionary or other changes holding out a bug fix release?

eustas:
Hopefully will release 1.0.10 (or is is 1.1?) right after NY.

andrerom:
> 1.0.10 (or is is 1.1?)

With new features in branch, v1.1 might be best to signal the changes 👍

PR #920, opened 2021-08-27
Migrate to github actions

eustas:
Travis + Appveyor are replaced with Github Actions.

Not all combinations are migrated to the initial configuration; corresponding TODOs added.

Drive-by: additional combinations uncovered minor portability problems -> fixed
Drive-by: remove no-longer used "script" files.


docs/decode.h.3
```diff
@@ -167,6 +171,35 @@ Partially done; should be called again with more input\&.
...
 .SH "Function Documentation"
 .PP 
+.SS "\fBBROTLI_BOOL\fP BrotliDecoderAttachDictionary (\fBBrotliDecoderState\fP * state, BrotliSharedDictionaryType type, size_t data_size, const uint8_t data[data_size])"
+
+.PP
+Adds LZ77 prefix dictionary, adds or replaces built-in static dictionary and transforms\&. Attached dictionary ownership is not transferred\&. Data provided to this method should be kept accessible until decoding is finished and decoder instance is destroyed\&.
+.PP
+\fBNote:\fP
+.RS 4
+Dictionaries could NOT be attached after actual decoding is started\&.
```
lvandeve, line 181: Dictionaries can NOT (or Dictionaries cannot) would be more correct here since it's not an action done in the past

lvandeve (approved):

docs/decode.h.3
eustas: Fixed. Thanks!

PR #921, opened 2021-08-30
doc: default windows size is 24

sebres:
Proposed patch fixes default window size (`-w`, `--lgwin`) in the documentation.

Since #656 it got increased to 24, see https://github.com/google/brotli/blob/19d86fb9a60aa7034d4981b69a5b656f5b90017e/c/tools/brotli.c#L86


sebres:
@googlebot I signed it!

eustas:
This file is generated from internal md file. Going to fix that one first.
Thanks for noticing the inconsistence.

eustas:
Integrated into #925

PR #925, opened 2021-09-06
Strip "./" in includes

eustas:
Drive-by: fix brotli.1 and brotli.md - see #921 (kudos goes to @sebres)


eustas:
@googlebot I consent.

c/common/shared_dictionary_internal.h
```diff
@@ -9,9 +9,9 @@
 #ifndef BROTLI_COMMON_SHARED_DICTIONARY_INTERNAL_H_
 #define BROTLI_COMMON_SHARED_DICTIONARY_INTERNAL_H_
 
-#include "./dictionary.h"
+#include "dictionary.h"
 #include <brotli/shared_dictionary.h>
```
lvandeve, line 13: can the includes be ordered by type here? first <> style includes, then "" style includes

c/dec/decode.c
```diff
@@ -16,10 +16,10 @@
 #include "../common/shared_dictionary_internal.h"
```
lvandeve, line 16: can the "../" cases also be eliminated?

lvandeve (approved):

c/dec/decode.c
eustas: Not sure. Current relative path scheme fits Bazel, CMake, Autotools, plain make and python disttools. Will check if we could reliably specify build root to all those build systems.

c/common/shared_dictionary_internal.h
eustas: Those are made via copybara translation; not sure if it is possible to sort lines in a reversible manner.
However, it might be possible to modify the internal codebase, in such a manner that library includes are separate from the regular ones.

PR #926, opened 2021-09-09
Make TODO/Author entries copybara-reversible.

eustas:
This change allows both automated copybara export and import.


eustas:
@googlebot I consent.

PR #929, opened 2021-09-14
Fix -Werror=vla-parameter errors with GCC 11.2.0

jbms:


c1728p9:
I'm having errors with GCC 11 as well when using brotli from edk2. This patch fixes the problem for me.

@eustas can this PR be merged, or is it waiting on something?

PR #931, opened 2021-10-10
Suppress cmake's CMP0048 warning

the-moisrex:
Not providing VERSION to "project" command will cause a warning.

Since this project's version is loaded from other files, this policy will help suppress the warning generated by cmake.
This policy is set because we can't provide "VERSION" in "project" command.
Use cmake --help-policy CMP0048 for more information


eustas:
Thanks.

PR #941, opened 2021-11-17
Print non-error (de)compression messages to stdout

Marcuse7:


eustas:
The purpose of printing everything in `stderr` is to make sure `stdout` does not get polluted with extra output.
There is a mode when output is pushed to `stdout` instead of file, so if everything else gets there, then data will be corrupted.

PR #946, opened 2021-12-29
python: Use underscore separated keys in setup.cfg

00-matt:
setuptools has deprecated dash-separated keys:

```
Usage of dash-separated 'build-base' will not be supported in future versions. Please use the underscore name 'build_base' instead
```


00-matt:
Whoops, I missed issue 911. Dupe.

PR #960, opened 2022-04-13
Provide length-limited decompress methods

CamW:
Provide length-limited decompress methods so that callers can control allocation sizes.

Motivation:
Limiting the size of buffer allocation can help mitigate the risks of compression bombs. In cases where these could be a legitimate attack vector, we should provide the library caller with the option of limiting the buffer size.

Changes:
 -  Add `decompressKnownLength` method to `Decoder.java` to allow for decompression of known-size payloads.
 -  Add `decompress` overload to `Decoder.java` which accepts a max size to allow for decompression of untrusted payloads with a maximum allowed size.
 -  Modify `DecoderJNI.java`, `decoder_jni.cc` and `decoder_jni.h` to allow users to specify the output buffer size.

Result:
Callers which believe that they have known-size payloads or want to limit decompressed data sizes can decompress those payloads in linear space complexity based on the expected/max-allowed size rather than unknown space complexity, based on the actual size.

### DIFF
--- a/java/org/brotli/wrapper/dec/BUILD
+++ b/java/org/brotli/wrapper/dec/BUILD
@@ -101,3 +101,16 @@ java_test(
     test_class = "org.brotli.wrapper.dec.CornerCasesTest",
     runtime_deps = [":test_lib"],
 )
+
+java_test(
+    name = "LengthLimitedTest",
+    size = "large",
+    data = [
+        ":brotli_jni",  # Bazel JNI workaround
+    ],
+    jvm_flags = [
+        "-DBROTLI_JNI_LIBRARY=$(location :brotli_jni)",
+    ],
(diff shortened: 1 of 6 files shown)


eustas:
Hello. Thank you for the PR.

I believe, we do not need changes on native (JNI) side. `BrotliDecoderTakeOutput` only provides access to the internal buffer. `size` parameter only affects how much output is considered consumed, not how much memory is used.
So the changes on JNI side are no-op memory-wise.

As a consequence, it is better to write a tiny proxy around BrotliInputStream/BrotliInputChannel that stops reading when there is more data than expected. This way decompressor will become an interchangeable part - you could plug JNI decoder, Java decoder, or even some other format decoder...

PR #966, opened 2022-05-09
Wrap interface include directories with BUILD_INTERFACE generator expression

madebr:
When exporting a CMake target using install(TARGETS) + install(EXPORT),
CMake requires all include directories to be clean of build system
directories.

https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_INCLUDE_DIRECTORIES.html

This change also allows use of brotli as a CMake subproject and
installing + exporting it.

Needed for https://github.com/libsdl-org/SDL_image/pull/189 (brotli is a dependency of libjxl)


CMakeLists.txt
```diff
@@ -197,7 +197,7 @@ foreach(lib IN LISTS BROTLI_SHARED_LIBS BROTLI_STATIC_LIBS)
   if(NOT BROTLI_EMSCRIPTEN)
     set_target_properties(${lib} PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
   endif()
-  set_property(TARGET ${lib} APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${BROTLI_INCLUDE_DIRS}")
+  set_property(TARGET ${lib} APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES "$<BUILD_INTERfACE:${BROTLI_INCLUDE_DIRS}>")
```
madebr, line 200:
```suggestion
  set_property(TARGET ${lib} APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES "$<BUILD_INTERFACE:${BROTLI_INCLUDE_DIRS}>")
```

eustas (approved):

PR #967, opened 2022-05-11
Update CMakeLists.txt

Prakash222:


eustas:
First we have to ramp up minimum CMake version.

eustas:
> CMake version 3.25.1 warns when the policy is not set and uses OLD behavior.

I.e. if we move explicit policy value we will roll back to "OLD" deprecated behavior. No go.

PR #975, opened 2022-08-10
Add *.d to gitignore

Zapa289:
Ignore build files generated by C build.


eustas:
Hello. Thanks for the PR.
`.d` extension has "overloads" - it is both D sources and dependencies files. So adding it to ignore-list is a bit controversial.

IIUC, you are building using autotools and CFLAGS="-MMD"?
Current autotools script/config is imperfect. It puts all the output near source files. Let's improve it, towards cleaner build... and add output folder to ignore-list.

PR #977, opened 2022-09-01
VERBOSE-patch-1 

znzjugod:
fix verbose mode in brotli.c


eustas:
Please give a better description. From patch I can only see it breakes "unbrotli" alias.

PR #978, opened 2022-09-07
Fix bootstrap version computing with custom bc

michal-josef-spacek:
When i have ~/.bc configuration file with content: scale=2
which is changing default behaviour (scale=0), bootstrap is not working.


eustas (approved):
Thanks!

PR #987, opened 2022-11-08
add pyproject.toml

### LINKED ISSUE #984: Add 3.11 wheels
As Python3.11 will be released later today, it would be truly great of 3.11 wheels could be added to the [latest release](https://pypi.org/project/Brotli/1.0.9/#files).

wordlesswind:
```
DEPRECATION: brotli is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
```

pypa/pip#8559

https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#fallback-behaviour

fix #984

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools>=40.8.0", "wheel"]
+build-backend = "setuptools.build_meta:__legacy__"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools>=40.8.0", "wheel"]
-build-backend = "setuptools.build_meta:__legacy__"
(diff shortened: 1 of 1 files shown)


anthrotype:
you only need this:

```toml
[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"
```

we don't want to use "setuptools.build_meta.__legacy__", and [wheel is not needed](https://github.com/pypa/pip/issues/8559#issuecomment-1312478695)

anthrotype (approved):
LGTM

anthrotype:
ping @szabadka

PR #988, opened 2022-12-18
CMake: ensure static libraries are still installed on Emscripten

kleisauke:
Similar to commit ce222e317e36aa362e83fc50c7a6226d238e03fd.

---

Context:
https://github.com/google/brotli/pull/655#issuecomment-1353141509

FWIW, the OSS-Fuzz build of Brotli seems to fail.
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#brotli

Let me know if you want a PR for that too (a trivial change, since that build script was probably based on `c/fuzz/test_fuzzer.sh`)


kleisauke:
For reference, this was tested with commit https://github.com/kleisauke/wasm-vips/commit/9083bab88e8c400bf39a9c04931bc761551ec97e.

eustas:
Thanks

PR #1005, opened 2023-01-22
avoid array allocation in Encoder.Mode enum

sullis:
Reference:

https://www.gamlor.info/wordpress/2017/08/javas-enum-values-hidden-allocations/


eustas:
Awesome finding! Will land soon.

eustas (approved):

PR #1006, opened 2023-02-02
fuzz: more complete fuzzing

catenacyber:
cc @eustas 

This PR adds a fuzz target for encoding/compression

And it also completes the decode fuzzer by adding support for attaching a dictionary, and setting parameters (BROTLI_DECODER_PARAM_LARGE_WINDOW)
This modifies a bit the format that is used (using the first bytes to get the options). Is it ok ? Or should it be a new target ?

What do you think of it ?
PS : no bugs found after one night of running

### DIFF
--- a/c/fuzz/decode_fuzzer.c
+++ b/c/fuzz/decode_fuzzer.c
@@ -8,19 +8,34 @@
 
 #include <brotli/decode.h>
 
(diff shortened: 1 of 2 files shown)


catenacyber:
friendly ping @eustas ?

PR #1021, opened 2023-06-18
Add support for clang-cl compiler

fhanau:
The bazel build does not currently work with the Windows-based `clang-cl` compiler. This fixes the build and warnings for me.
In addition to the compiler warnings themselves, the flags `--pedantic-errors` and `-lm` are not supported by clang-cl and have been removed. clang-cl reports many warnings here and we may expect additional ones in future versions, so just removing `-Werror` and `--pedantic-errors` would also be an option, although the set of disabled warnings looks sufficient for now.


sezero:
clang-cl treats `-Wall` as `-Weverything` which is very loud. Suggest using `/W4` which clang-cl treats as `-Wall -Wextra`

fhanau:
> clang-cl treats `-Wall` as `-Weverything` which is very loud. Suggest using `/W4` which clang-cl treats as `-Wall -Wextra`

Thank you @sezero! I updated the PR to use `/W4`, looks much cleaner now.

eustas (approved):

fhanau:
Unfortunately Windows CI builds are failing due to https://github.com/bazelbuild/bazel/issues/18592. This has not been fixed in bazel yet, but a workaround is to downgrade the GitHub Actions runner image to `windows-2019`. Note that this will affect not only this but all other pull requests.
@eustas can you make that change? It's inconvenient, but looks like the easiest workaround. I could also try changing it in it PR but I'm not sure if I have the permissions to do so.

eustas:
Hoi. Could you rebase, please. Fixed build, and currently CI is green

fhanau:
> Hoi. Could you rebase, please. Fixed build, and currently CI is green

Thank you, I rebased the PR.

fhanau:
Builds are now passing, feel free to merge the PR.

PR #1022, opened 2023-06-19
Add loongarch64 support

loongson-zn:
define BROTLI_TARGET_64_BITS is 1 on loongarch64 platform

### DIFF
--- a/c/common/platform.h
+++ b/c/common/platform.h
@@ -208,8 +208,13 @@ OR:
 #define BROTLI_TARGET_RISCV64
 #endif
 
+#if defined(__loongarch_lp64)
+#define BROTLI_TARGET_LOONGARCH64
+#endif
+
 #if defined(BROTLI_TARGET_X64) || defined(BROTLI_TARGET_ARMV8_64) || \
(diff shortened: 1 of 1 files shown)


eustas (approved):

eustas:
Hi. Thanks for PR.

Is there a way to compile for this arch in github actions?

loongson-zn:
I verified it on my personal computer,  but I believe on [Gentoo](https://www.gentoo.org/downloads/#loong) or other upcoming systems (openEuler and Debian) will be successful. 

```
69/70 Test #69: compatibility/tests/testdata/xyzzy.compressed .................   Passed    0.01 sec
      Start 70: compatibility/tests/testdata/zeros.compressed
70/70 Test #70: compatibility/tests/testdata/zeros.compressed .................   Passed    0.01 sec

100% tests passed, 0 tests failed out of 70

Total Test time (real) =   5.35 sec
[zhangna@localhost build]$ arch
loongarch64

```

loongson-zn:
> Thanks for testing, but I'm more worried that future changes won't break build for this arch...

I don't think this problem will occur, at least I can guarantee that there won't be any problem with the LongArch code. If you still have concerns, I will continue to follow up

eustas:
Landed

PR #1089, opened 2023-09-27
CMake: install relocatable shared libs for macOS

SpaceIm:
Remove hardcoded `CMAKE_INSTALL_NAME_DIR`, so that it's `@rpath` by default (default CMake behavior when cmake_minimum_required() is at least 3.0). `set(CMAKE_MACOS_RPATH TRUE)` is useless, it's the default value for CMake if cmake_minimum_required() is at least 3.0 as well, no need to hardcode this value.

If someone wants to change this behavior for a specific reason, it can be injected externally, no need to patch build files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,8 +112,6 @@ elseif(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
   add_definitions(-DOS_FREEBSD)
 elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   add_definitions(-DOS_MACOSX)
-  set(CMAKE_MACOS_RPATH TRUE)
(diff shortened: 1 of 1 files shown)


eustas:
Hello. I'll prepare a middle-ground PR:
 - remove `set(CMAKE_MACOS_RPATH`
 - remove `set(CMAKE_INSTALL_NAME_DIR`
 - specify default `INSTALL_NAME` if `CMAKE_INSTALL_NAME_DIR` is not defined

SpaceIm:
> specify default INSTALL_NAME if CMAKE_INSTALL_NAME_DIR is not defined

Why?

eustas:
Because that is what embedders of library do. Why not make their life easier?

eustas (approved):

SpaceIm:
It makes life of package managers harder, because it changes default CMake behavior.

eustas (approved):

PR #1084, opened 2023-09-17
CMakeLists.txt: use CMAKE_INSTALL_FULL_* directories to be friendlier to distributions

trofi:

PR #1096, opened 2023-11-01
Add SHARE_INSTALL_PREFIX to building with cmake

TedLyngmo:
Without `-DSHARE_INSTALL_PREFIX=...` it will try to install the `man` pages in the root directory.


TedLyngmo:
I just noticed a much better PR solving the same thing here: https://github.com/google/brotli/pull/1084

PR #1104, opened 2023-12-28
Java wrapper Encoder wrong length 

hyperxpro:
Problem:

Currently, the Java wrapper Encoder is passing byte array length for finishing the encoding of data. This is ideal when the encoder is used with offset 0 and full length is passed but for cases where offset is passed and length is also specified, it leads to encoded data corruption.

Changes:
Instead of passing byte array length to the `push` method, we will now pass `length` directly.


hyperxpro:
@eustas PTAL

eustas:
Looking into

eustas:
Indeed! Let me create a testcase for that first (likely will be ready tomorrow).

eustas (approved):

eustas (approved):

PR #1163, opened 2024-04-18
print help to stdout

tansy:
According to [GNU coding standards](https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dhelp) \`--help' should be printed to stdout. It makes sense as long help screen can be paged (\`$ prog --help | more`), searched and so on.
There is an issue with the paging here because when paged with less(1) it makes black screen and even leaving it is difficult. I understand that is the issue but I made it to acknowledge the problem.


tansy:
I'm not going to jump through the hoops to commit oneliners, that are technically trivial, even if I, personally think they are important.
Also who do I contact about this 'failed invocation' script?

PR #1172, opened 2024-05-26
Add cmake version and target

bmarques1995:
This commit will add a cmake target, which is possible to use the command `find_package(Brotli)` in cmake


bmarques1995:
> Packaging and distribution of Brotli is more difficult than it should. This PR helps. Thanks.

I can't run the workflows, they are stuck
Captura de tela 2024-09-06 004937

bmarques1995 (commented):
Let's test it

PR #1185, opened 2024-08-07
cbrotli: add pkg-config directive

dloebl:
Adds the pkg-config directive for `libbrotlicommon`, `libbrotlidec` and `libbrotlienc`. Without this change, building `cbrotli` may fail on some platforms because the include directories are not set correctly:
```
github.com/google/brotli/go/cbrotli
# github.com/google/brotli/go/cbrotli
vendor/github.com/google/brotli/go/cbrotli/reader.go:13:10: fatal error: 'brotli/decode.h' file not found
#include <brotli/decode.h>
         ^~~~~~~~~~~~~~~~~
1 error generated.
```

### DIFF
--- a/go/cbrotli/cgo.go
+++ b/go/cbrotli/cgo.go
@@ -7,7 +7,5 @@ package cbrotli
 
 // Inform golang build system that it should link brotli libraries.
 
-// #cgo LDFLAGS: -lbrotlicommon
-// #cgo LDFLAGS: -lbrotlidec
-// #cgo LDFLAGS: -lbrotlienc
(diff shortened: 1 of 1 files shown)


dloebl:
Hey @eustas, sorry to ping you on this. I just wanted to ask if you are still maintaining `google/brotli`?
Having this fix in a release would really help us, because as of now, we always have to apply this patch to get `cbrotli` working properly

eustas:
Looks good. Thanks! 
I'll explore the topic and land this PR soon.

eustas (approved):

PR #1203, opened 2024-10-01
Add type stubs for python module

dbowring:


eustas:
Hi. It seems that #1206 contains more thorough stub.

PR #1218, opened 2024-11-11
Fix Java Bazel build

eustas:


eustas:
Still windows tests are failing

PR #1230, opened 2024-12-20
Remove old Python 3.x PyPI trove classifiers

kurtmckee:
Because the `python_requires` argument is not used, there are no restrictions on what Python versions are supported. It therefore doesn't make sense to list a small subset of EOL Python 3.x versions.


kurtmckee:
Closing for lack of response.

PR #1234, opened 2025-01-06
Add support for limiting output size in Python streaming decompression, take 2

robryk:
This is a fixed version of #1201


eustas (approved):

hamidnivora:
Yes

eustas (approved):

Dreamsorcerer:
@robryk I'm struggling to find any documentation for this. What is the correct way to read the full body while using this parameter?

e.g. (In Python)
zlib:
```
output = d.decompress(inp, max_length)
while d.unconsumed_tail:
    output += d.decompress(d.unconsumed_tail, max_length)
```
zstd:
```
output = d.decompress(inp, max_length)
while not d.eof and not d.needs_input:
    output += d.decompress(b"", max_length)
```

Obviously concatenating to `output` defeats the point, I'm just trying to keep the examples simple.

Dreamsorcerer:
Answering my own question, there doesn't appear to be any API to know whether more output is available, the only way to check is to keep retrying until there's no output, so the code would look like:
```
output = d.process(inp, max_length)
while out := d.process(b"", max_length):
    output += out
```

Which seems a bit awkward to me and means you'll likely need an extra call at the end to know you've consumed all the data.

There's also a .can_accept_more_data() which returns False if .process() will throw an exception when passed something other than `b""`. This strikes me as a little weird too given that the initial call will accept any amount of input, so rejecting further input seems pointless to me.

PR #1240, opened 2025-01-09
Update bro.py: Fix undefined name in Python code

cclauss:
`outfile` is used before it is defined on the following line so use `options.outfile` instead.

% `ruff check`
```
Error: python/bro.py:168:64: F821 Undefined name `outfile`
```


eustas:
Thanks!

Issue #1245, opened 2025-01-18
Apple Frameworks


Treata11:
Greetings,

I noticed that the project's CMake minimum version is set to 3.15, which is sufficient for Apple platform builds with CMake's official toolchain.
https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#id27

But I took a look at the source & couldn't find the [Framework properties](https://cmake.org/cmake/help/latest/prop_tgt/FRAMEWORK.html#prop_tgt:FRAMEWORK) set in project's configs...
I'd be happy to submit a PR to address this.

PR #1246, opened 2025-01-18
Add support for Apple-Framework builds

Treata11:
Greetings, 

I introduced the required properties in `CMakeLists.txt` to enable cross-compilation for iOS-derived platforms as explained in the [CMake documentation](https://cmake.org/cmake/help/latest/prop_tgt/FRAMEWORK.html#prop_tgt:FRAMEWORK).

Binaries  of `brotlicommon.framework`, `brotlidec.framework` & `brotlienc.framework` targets will be generated in the `build` directory with the following commands:
```bash
cmake -S. -B build -DBUILD_FRAMEWORK=TRUE -DCMAKE_SYSTEM_NAME=iOS -DBROTLI_BUNDLED_MODE=ON

cmake --build build --target install --config Release
```

It's also worth noting that I included the `BROTLI_PUBLIC_HDRS` files for the targets above to ensure they are included in the built `framework` bundles.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,10 @@ project(brotli C)
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 set(BROTLI_BUILD_TOOLS ON CACHE BOOL "Build/install CLI tools")
 
+if(APPLE)
+  option(BUILD_FRAMEWORK "Build as Apple Frameworks" OFF)
+endif()
+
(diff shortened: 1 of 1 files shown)


Treata11:
Addresses issue #1245

Treata11:
@eustas, Changes above are sufficient to build for iOS-derived platforms & their simulators.
Separate framework bundles are generated for each target: 
**(`brotlicommon.framework`, `brotlidec.framework` & `brotlienc.framework`)**.

PR #1255, opened 2025-02-26
Fix ISO C build breakage.

robinwatts:
ISO C prohibits inline declarations of variables. Move the declaration to the start of the block.

### DIFF
--- a/c/dec/decode.c
+++ b/c/dec/decode.c
@@ -1410,6 +1410,7 @@ static BROTLI_BOOL BROTLI_NOINLINE BrotliEnsureRingBuffer(
 static BrotliDecoderErrorCode BROTLI_NOINLINE
 SkipMetadataBlock(BrotliDecoderState* s) {
   BrotliBitReader* br = &s->br;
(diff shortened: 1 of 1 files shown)


c/dec/decode.c
```diff
@@ -1420,7 +1421,7 @@ SkipMetadataBlock(BrotliDecoderState* s) {
   /* Drain accumulator. */
   if (BrotliGetAvailableBits(br) >= 8) {
     uint8_t buffer[8];
-    int nbytes = (int)(BrotliGetAvailableBits(br)) >> 3;
+    nbytes = (int)(BrotliGetAvailableBits(br)) >> 3;
```
a1batross, line 1424: I'm just passing by, but you don't need to move this declaration, as it's already at the start of it's own block.

robinwatts: Indeed. But by moving the other declaration, this one shadows it.

It doesn't affect the correctness of the code, but it's "bad form" and some static analysers might whinge about it.

PR #1286, opened 2025-06-18
remove google3 from opens-source imports

victorvianna:


victorvianna:
Was already addressed by copybara here https://github.com/google/brotli/commit/418237cff452c5da88934e6343d9cd3f66d99347

PR #1367, opened 2025-10-27
cmake: fix man page install directories

vszakats:
To not use a prefix-less absolute path with
`cmake --install <d> --prefix <path>`.

According to CMake docs about `_FULL_` variables:
"These variables shouldn't be used in install() commands as they do
not work with the cmake --install command's --prefix option, or with
the cpack installer generators."

Ref: https://cmake.org/cmake/help/v4.2/module/GNUInstallDirs.html

Follow-up to cff58032160406c170a3dc319ad6e7b288cafcc0 #1084

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -384,11 +384,11 @@ endif()  # BROTLI_BUNDLED_MODE
 
 if (BROTLI_BUILD_TOOLS)
   install(FILES "docs/brotli.1"
-    DESTINATION "${CMAKE_INSTALL_FULL_MANDIR}/man1")
+    DESTINATION "${CMAKE_INSTALL_MANDIR}/man1")
 endif()
 
 install(FILES docs/constants.h.3 docs/decode.h.3 docs/encode.h.3 docs/types.h.3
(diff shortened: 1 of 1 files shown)


vszakats:
I'm not ready to sign a CLA, please use this as-is.

The issue causes a breakage after bumping to 1.2.0 (from 1.1.0) as visible in this build:
https://github.com/curl/curl-for-win/actions/runs/18853487748/job/53795465550
```
[38/38] Linking C static library libbrotlienc.a
+ cmake --install _a64-mac-sys-bld --prefix _a64-mac-sys/usr
-- Install configuration: "Release"
CMake Error at _a64-mac-sys-bld/cmake_install.cmake:89 (file):
  file cannot create directory: /usr/share/man/man3.  Maybe need
  administrative privileges.
```

Patch confirmed working:
https://github.com/curl/curl-for-win/actions/runs/18853837259/job/53796629171

In different envs, or when run with sudo, this may cause overwriting
system man pages with the one from this build, also when attempting
to install to a non-system prefix, with the cmake `--prefix` option.

vszakats:
Closing in favor of #1371. Though feel free to reopen, or merge anyway.

PR #1368, opened 2025-10-29
Revert "explicitly specify model for relocatable variables"

xry111:
This reverts commit 103b25fb1e38ee765c827ce470c85a4d6541df12.

That breaks compilation on at least LoongArch and IA-64 with latest GCC because LoongArch GCC does not recognize "small" and IA-64 GCC accepts an identifier (instead of string literal).

Per the ChangeLog it seems the change was for "supporting binaries larger than 2 GiB" but I cannot see how it will work reliably.  If a binary will be larger than the limit of the default code model, the entire binary (i.e. each relocatable object file that would consist the binary) must be compiled with a larger code model because the linker does not guarantee any specific ordering of data sections from the inputs, so there will be a good chance that other data objects (not only the table) can be out of the 2 GiB range.  And, the definition of code models varies between targets, so "small" may be not suitable for some of them even if it's accepted.

So to me for now we should revert it.  If we really need to override the code model for some targets we should add -mcmodel= options (depending on the target) into the build system instead.


eustas:
Hi. Thanks for explanation. Let's stay on the middle-ground: disable BROTLI_MODEL macro for LoongArch and IA-64. I'll prepare PR for that.

PR #1379, opened 2025-11-07
Fix method name in python Decompressor.can_accept_more_data docstring

Cycloctane:
It should be `Decompressor.process()`.


eustas:
Thanks!

PR #1392, opened 2025-11-20
Add ruff rule PERF for performance

cclauss:
% `ruff check --extend-select=C4,C90,PERF,RET,SIM,W`

% `ruff rule PERF401`
# manual-list-comprehension (PERF401)

Derived from the **Perflint** linter.

Fix is sometimes available.

## What it does
Checks for `for` loops that can be replaced by a list comprehension.

## Why is this bad?
When creating a transformed list from an existing list using a for-loop,
prefer a list comprehension. List comprehensions are more readable and
more performant.

Using the below as an example, the list comprehension is ~10% faster on
Python 3.11, and ~25% faster on Python 3.10.

Note that, as with all `perflint` rules, this is only intended as a
micro-optimization, and will have a negligible impact on performance in
most cases.

## Example
```python
original = list(range(10000))
filtered = []
for i in original:
    if i % 2:
        filtered.append(i)
```

Use instead:
```python
original = list(range(10000))
filtered = [x for x in original if x % 2]
```

If you're appending to an existing list, use the `extend` method instead:
```python
original = list(range(10000))
filtered.extend(x for x in original if x % 2)
```

### DIFF
--- a/c/enc/backward_references_hq.c
+++ b/c/enc/backward_references_hq.c
@@ -433,7 +433,7 @@ static size_t UpdateNodes(
   const CompoundDictionary* addon = &params->dictionary.compound;
   size_t gap = addon->total_size;
 
-  BROTLI_DCHECK(cur_ix_masked + max_len <= ringbuffer_mask);
+  BROTLI_DCHECK(cur_ix_masked + max_len <= ringbuffer_mask + 1);
 
   EvaluateNode(block_start + stream_offset, pos, max_backward_limit, gap,
       starting_dist_cache, model, queue, nodes);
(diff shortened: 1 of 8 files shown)


eustas:
Cool. Thanks.

cclauss:
How do pull requests progress if ___required___ are not run?

Should ___required___ tests be run on all pull requests?  Or do maintainers just merge without the ___required___ tests being run?

cclauss:
Thanks for the review.  Three ___required___ GitHub Actions workflows need maintainer approval below to run.

eustas:
No worries about workflows - landing of PR in google/brotli is a bit more complicated than just almost everywhere in GH. You might have noticed that it is "copybara-service" pushes commits...

eustas (approved):

PR #1398, opened 2025-11-30
fix 'const' qualifier in c\dec\bit_reader.c

KilianKegel:
different 'const' qualifier in c\dec\bit_reader.c and qualifier in c\dec\bit_reader.h

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -133,11 +133,22 @@ jobs:
             c_compiler: clang
             cxx_compiler: clang++
 
-          - name: python3.10-win
+          - name: python3.14:clang
             build_system: python
-            python_version: "3.10"
-            # TODO: investigate why win-builds can't run tests
-            os: windows-2022
+            python_version: "3.14"
+            c_compiler: clang
+            cxx_compiler: clang++
+
+          - name: python3.14t:clang
+            build_system: python
+            python_version: "3.14t"
+            c_compiler: clang
+            cxx_compiler: clang++
+
+          - name: python3.14-win
+            build_system: python
+            python_version: "3.14"
+            os: windows-latest
 
           - name: maven
             build_system: maven
@@ -331,8 +342,7 @@ jobs:
       if: ${{ matrix.build_system == 'python' }}
       run: |
         python -VV
-        python -c "import sys; sys.exit('Invalid python version') if '.'.join(map(str,sys.version_info[0:2])) != '${{ matrix.python_version }}' else True"
-        pip install setuptools==51.3.3 pytest
+        pip install "setuptools>=70.0.0" pytest
         python setup.py build_ext --inplace
         pytest ./python/tests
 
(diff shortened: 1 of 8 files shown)


jyrkialakuijala:
why is the current state of things a problem?

KilianKegel:
Hi Jyrki

I am an EDK2/BIOS developer for fun and profit.
I have own projects at my github site based on original Intel Tianocore EDK2.
E.g. The UEFI SHELL: https://github.com/KilianKegel/UEFI-SHELL
While updating to the latest Intel Tianocore EDK2 202511 it appears that at startup (in my case LAUNCH.BAT)
can not translate brotli component, due to a warning.
https://github.com/tianocore/edk2/tree/master/BaseTools/Source/C/BrotliCompress

You can checkout the above (first) link. It builds without any additional installation and you can see build break LAUNCH.bat.

The 'const' keyword is different between definition and declaration.
It is obviously wrong!

Hopefully EDK2/Tianocore will use the fix in the next EDK2 release.

Thanks a lot,
Kilian
________________________________

jyrkialakuijala:
ISO/IEC 9899:1999 (C99 Standard), Section 6.7.5.3, Paragraph 15. (ok, I don't really know the C99 standard, I just asked gemini) supposedly says that the prototype of the function may leave the const away from constant pointers even if the implementation uses a constant pointer. It seems to me that displaying a warning is a compiler error.

Do you have a recommendation on how to proceed given that what I wrote is most likely true?

KilianKegel:
Hi Jurki,
in bit_reader.c and bit_reader.h function definition and declaration must be the same. It’s different currently regarding the ‚const‘ qualifier. It’s obviously wrong! In my build it generates a warning that breaks the build. I can send you screenshot later in the evening.
Thanks, Kilian

________________________________

KilianKegel:
[cid:c6795975-362a-42aa-ae6e-98a3bc6930dc]
Hi Jyrki, my error message above.

Thanks for your help,
Kilian

________________________________

jyrkialakuijala:
I wasn't able to open the error msg.

KilianKegel:
Uploaded to m github account
https://github.com/KilianKegel/brotli_build_error_in_EDK2/blob/main/image.png

Regards,
Kilian
________________________________

KilianKegel:
Hi Jyrki,
may I ask when the master branch is updated with the fix?

Thanks,
Kilian

eustas:
Hi. Change mostly looks good to me, but could we do vice-versa: propagate `const` qualifier to header?

eustas:
Please ignore my last comment - `const` in header was dropped in 2023, because it is meaningless.

eustas:
Both @jyrkialakuijala and MSVC are right. It is okay-ish in C99, but we are aiming at C89, where `const` qualifier changes type "compatibility": https://stackoverflow.com/questions/18524336/is-visual-studios-c4028-warning-formal-parameter-different-from-declaration-s

PR #1406, opened 2025-12-10
Fix Emscripten detection false positives and unsuppressible status me…

mannewalis:
…ssages

Issue 1: False positive Emscripten detection

The check_c_source_compiles() test incorrectly detected Emscripten when building with non-Emscripten WASM toolchains (e.g., WASI SDK). In cross- compilation scenarios, CMake creates a static library instead of an executable for compile tests. An empty source file (when __EMSCRIPTEN__ is not defined) successfully archives into a static library, causing the test to pass when it should fail.

Resolution: Use the EMSCRIPTEN CMake variable set by Emscripten's official toolchain file, which is the recommended detection method. Remove the unreliable check_c_source_compiles() fallback.

Issue 2: Unsuppressible status messages

The message() calls lacked a type argument, making them always print regardless of CMake's log level settings.

Resolution: Add STATUS type to message() calls so they can be suppressed with CMake's --log-level option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,17 +17,23 @@ else()
   message(STATUS "Build type is '${CMAKE_BUILD_TYPE}'")
 endif()
 
-include(CheckCSourceCompiles)
-check_c_source_compiles(
-  "#if defined(__EMSCRIPTEN__)
-   int main() {return 0;}
-   #endif"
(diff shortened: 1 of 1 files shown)


eustas (approved):

eustas:
Thanks!

PR #1412, opened 2026-01-20
Support building w/ Bazel 9

### LINKED ISSUE #1405: brotli doesn't work with bazel 9.0
Tried with both 1.1.0 and 1.2.0. 
Looks like this needs a module decency on `rules_cc` and to load `cc_library` from there.

```
ERROR: Traceback (most recent call last):
        File "C:/dev/_bazel/nthlhj2p/external/+http_archive+com_google_brotli/BUILD.bazel", line 91, column 11, in <toplevel>
                cc_library(
        File "/virtual_builtins_bzl/bazel/exports.bzl", line 40, column 9, in _removed_rule_failure
Error in fail:
         This rule has been removed from Bazel. Please add a `load()` statement for it.
         This can also be done automatically by running:
         buildifier --lint=fix <path-to-BUILD-or-bzl-file>
```

robinlinden:
The builtin cc rules have been removed, and now have to be loaded from rules_cc.

Resolves #1405

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -16,7 +16,7 @@ README.md !export-ignore
 SECURITY.md !export-ignore
 setup.cfg !export-ignore
 setup.py !export-ignore
-WORKSPACE.bazel !export-ignore
+MODULE.bazel !export-ignore
 
 # Add sources
 c !export-ignore
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,6 +1,8 @@
 # Description:
 #   Brotli is a generic-purpose lossless compression algorithm.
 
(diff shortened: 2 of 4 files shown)


stevenlr:
I've also noticed that the MODULE.bazel file was not exported in the source archive. Perhaps this could be a good opportunity to fix that too.

In the `.gitattributes` files, the line `WORKSPACE.bazel !export-ignore` should be replaced with `MODULE.bazel !export-ignore`.

robinlinden:
@stevenlr Nice catch! Added a commit including `MODULE.bazel` in the source archives. :)

eustas:
Hi. Thanks for the PR. I believe Bazel 9 build has been fixed recently.

PR #1413, opened 2026-01-21
java: add explicit null check in BrotliInputStream.read

leofernandesmo:
- **Problem**: BrotliInputStream.read(byte[], int, int) currently dereferences destBuffer.length during bounds checking, so passing a null buffer triggers a NullPointerException from the dereference rather than an explicit parameter check.

- **Why change**: InputStream.read(byte[], int, int) specifies NullPointerException when the buffer argument is null, so this change preserves the expected exception type while making the failure mode explicit and easier to diagnose.

Repro:
```
InputStream in = InputStream.nullInputStream();
BrotliInputStream br = new BrotliInputStream(in);
br.read((byte[]) null, 0, 1);

```

- **Fix**: add early destBuffer == null check and throw NullPointerException("destBuffer is null").
- **CLA note**: “I understand this repo requires the Google CLA and will complete it if/when requested.”

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,7 +3,7 @@ tests/testdata/* binary
 # Exclude everything
 **/** export-ignore
 
-# Add top-level files
+# Allowlist some filenames
 .bazelignore !export-ignore
 BUILD.bazel !export-ignore
 CHANGELOG.md !export-ignore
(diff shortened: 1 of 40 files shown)


leofernandesmo:
I already signed the CLA

PR #1431, opened 2026-03-23
[Python] Add output_buffer_limit to brotli.decompress()

scovetta:
This PR adds `output_buffer_limit` to the `brotli.decompess()` function, to echo similar functionality in Decompressor.process(), enabling users to more safely decompress untrusted compressed data.

The default `output_buffer_limit` value of 0 means "unlimited", which preserves backwards compatibility. When set, `decompress()` raises `brotli.error` if the decompressed output would exceed the specified limit.

### DIFF
--- a/python/_brotli.c
+++ b/python/_brotli.c
@@ -72,6 +72,8 @@ static const char kDecompressSinkError[] =
     "brotli: decoder process called with data when 'can_accept_more_data()' is "
     "False";
 static const char kDecompressError[] = "brotli: decoder failed";
+static const char kDecompressOutputLimitError[] =
+    "brotli: decompressed output exceeds output_buffer_limit";
 
 /* clang-format off */
 PyDoc_STRVAR(brotli_error_doc,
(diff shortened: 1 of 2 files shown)


eustas:
Hi. Thanks for the PR.

Where this new API will be used? I'm asking, because unlike decoding, for encoding guarantees the output size limit: see `BrotliEncoderMaxCompressedSize`.

scovetta:
@eustas This shouldn't be a new API, just a new (optional) parameter on the publicly exposed `brotli.decompress()` module-level function, so if an application could just do something like:

```python
import brotli
import requests

res = requests.get('<url>')
- data = brotli.decompress(res.content)
+ data = brotli.decompress(res.content, output_buffer_limit=5_000_000)
```

The scenario we're trying to avoid is where the contents of `<url>` are absurdly large, or any other scenario where the bytes being decompressed aren't already trusted.

eustas:
Sorry for the confusion, I've misread (compression instead of decompresssion).

Will take a deep look soon-ish.

Again, thanks for the PR.

eustas:
So, current implementation allows situations when output is larger than requested limit.

Should we pass limit to "BufferInit", where it can remember the value and respect it in "BufferGrow"?

PR #1458, opened 2026-04-15
Java code cleanup for the dec package

rbri:
* use try-with-resources

### DIFF
--- a/java/org/brotli/dec/CompoundDictionaryTest.java
+++ b/java/org/brotli/dec/CompoundDictionaryTest.java
@@ -10,6 +10,7 @@ import static org.junit.Assert.assertEquals;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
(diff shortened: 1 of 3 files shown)


eustas:
Hello. Thanks for your PR.

I do not like useless things too, but there is a hidden agenda. Most of those are used to make transpiler life easier:
 * type casts are good for translating to golang (promotion rules could be implemented in transpiler, but that needs time)
 * "else" branches are used for conditional compilation (please note that condition is kind-of `constexpr`) plus inlining (again, otherwise we would need code in transpiler to eliminate unreachable things)

Originally project aimed to support older versions of Java down to 5, thus no try-with-resources and other features. As for now, target could be at least Java 8, so introducing try-with-resources is fine.

Again, thank you!

eustas:
As for array-copy: it is in cold path, it is not "portable", it is short. Overall - simple for loop will be faster here.

rbri:
@eustas thanks for the fast feedback, not sure what to do next - should i close this and make a new one with only the try-with resources stuff?

eustas:
It is fine to prune this PR.

rbri:
@eustas next try ;-)

eustas (approved):

eustas:
Awesome. Thanks.

Issue #1462, opened 2026-04-19
[Bug] High: CLI -f follows symlinked output path and overwrites target file


parasol-aser:
Priority: High

Affected component: `brotli(1)` CLI output file creation in `c/tools/brotli.c` (`OpenOutputFile()`)

Root cause:
`OpenOutputFile()` opens the output path with `open(output_path, O_CREAT | (force ? 0 : O_EXCL) | O_WRONLY | O_TRUNC, ...)`. When `-f` / `--force` is used, `O_EXCL` is removed and the code still does not set `O_NOFOLLOW`. A pre-existing symlink at the output path is therefore followed, and the symlink target is truncated and rewritten with attacker-controlled decompressed content.

Exact reproduction steps:
1. Use the current `master` HEAD that was verified locally (`ab685df`) and build the CLI (`build-asan/brotli` was used for verification).
2. Run the deterministic repro helper from the verification artifacts:

```bash
timeout 180s tests/regression/t02/repro_symlink_overwrite.sh build-asan/brotli
```

3. Observe the success output:

```text
T-02 OK: 21-byte seed overwrote symlink target with crafted plaintext
```

4. The helper creates `source.txt -> target.txt`, then runs `brotli -d -f source.txt.br`. After the run, `target.txt` contains `ATTACKER_CONTENT`, showing that the CLI followed the symlink and overwrote the target file instead of refusing the path.

Relevant file paths:
- `c/tools/brotli.c:795-816` (`OpenOutputFile()`)
- In particular `c/tools/brotli.c:803-804` (output opened without `O_NOFOLLOW`; `-f` removes `O_EXCL`)
- Verification repro assets: `tests/regression/t02/repro_symlink_overwrite.sh`, `tests/regression/t02/source.txt.br`, `tests/regression/t02/source.txt`, `tests/regression/t02/target.txt`

Suggested fix direction:
- Add `O_NOFOLLOW` to the output `open()` flags unconditionally.
- Keep `-f` scoped to overwriting an existing regular file; it should not implicitly allow following symlinks.
- If `open()` fails with `ELOOP`, emit a clear diagnostic so the refusal is obvious to users.
- On platforms without `O_NOFOLLOW`, use a final-component symlink check before opening as defense in depth.

PR #1463, opened 2026-04-19
Fix CopyStat TOCTOU in brotli CLI (#1461)

### LINKED ISSUE #1461: [Bug] High: CLI CopyStat uses path-based chmod/chown after closing output file
Priority: High

Affected component: `brotli(1)` CLI output metadata handling in `c/tools/brotli.c` (`CopyStat()` / `CloseFiles()`)

Root cause:
`CloseFiles()` closes `context->fout` first and only then calls `CopyStat()` on `context->current_output_path`. `CopyStat()` uses path-based `chmod()` and `chown()` on that pathname instead of fd-based metadata updates. An attacker who can replace the output pathname after `fclose()` can swap it to a symlink and redirect the metadata operations to an arbitrary target.

Exact reproduction steps:
1. Use the current `master` HEAD that was verified locally (`ab685df`) and build the CLI (`build-asan/brotli` was used for verification).
2. Run the deterministic single-shot repro helper from the verification artifacts:

```bash
timeout 180s tests/regression/t01/repro_copystat_swap.sh \
  build-asan/brotli \
  tests/regression/t01/libfclose_swap.so
```

3. Observe the success output:

```text
T-01 OK: target mode flipped to 644 via post-close CopyStat path
```

4. The helper swaps the just-closed output path to a symlink targeting `tests/regression/t01/target.txt`. After the run, the target file's mode changes from `0600` to `0644` while its content hash remains unchanged, proving the metadata write landed on the symlink target rather than the original output file.

Relevant file paths:
- `c/tools/brotli.c:855-879` (`CopyStat()` uses `chmod()` / `chown()` on `output_path`)
- `c/tools/brotli.c:1030-1045` (`CloseFiles()` calls `fclose()` before `CopyStat()`)
- `c/tools/brotli.c:1475` (`copy_stat` defaults to enabled)
- Verification repro assets: `tests/regression/t01/repro_copystat_swap.sh`, `tests/regression/t01/fclose_swap.c`, `tests/regression/t01/libfclose_swap.so`, `tests/regression/t01/plain.bin.br`, `tests/regression/t01/target.txt`

Suggested fix direction:
- Move metadata copying to happen before `fclose()` while the output fd is still available.
- Replace path-based `chmod()` / `chown()` with `fchmod()` / `fchown()` on `fileno(context->fout)`.
- Skip metadata copy when there is no real output fd (for example stdout).
- Keep this aligned with symlink hardening in `OpenOutputFile()` so the initial open also refuses symlinks.

parasol-aser:
## Summary

Fixes google/brotli#1461.

`CloseFiles()` in `c/tools/brotli.c` previously called `fclose(context->fout)` **before** invoking `CopyStat()` on the output pathname. `CopyStat()` then used path-based `chmod()` and `chown()`, which gave an attacker with write access to the output directory a window between the close and the metadata syscalls to swap the output pathname to a symlink and redirect the metadata writes to an arbitrary target.

Close the race by doing the metadata copy while the output file is still open, and by switching to fd-based syscalls on `fileno(fout)`:

- `CopyStat()` now takes `FILE* fout` and calls `fchmod()` / `fchown()` on the underlying fd.
- `CopyTimeStat()` uses `futimens(fd, ...)` in the `HAVE_UTIMENSAT` branch; the legacy `utime()` fallback is preserved for platforms without it.
- `CloseFiles()` invokes `CopyStat()` **before** `fclose()`, and skips it when `current_output_path == NULL` (stdout), matching the guidance in the issue.
- The Windows no-op shims are updated from `chmod` / `chown` to `fchmod` / `fchown` to keep the file compiling on `_WIN32`.
- An `fflush(fout)` is added before the timestamp copy so stdio-buffered output does not clobber the atime/mtime the helper just applied during `fclose()`'s write-back.

The old `/* TOCTOU violation ... */` comment is deleted — the violation is gone.

No CMake / Bazel / build-system changes are required; the fix uses only POSIX symbols already available in the existing guarded blocks, and stays within the C89 compatibility constraint from `CONTRIBUTING.md`.

## Test plan

Regression assets are checked in under `tests/regression/t01/`.

- [x] `tests/regression/t01/repro_copystat_swap.sh` — deterministic attack reproducer using an `LD_PRELOAD` hook on `fclose()` that swaps the just-closed output path to a symlink. Against pre-fix `brotli`, the script exits 0 and prints `T-01 OK: target mode flipped to 644 via post-close CopyStat path`. Against the patched binary, the attack no longer lands.
- [x] `tests/regression/t01/test_copystat_swap_fixed.sh` / `test_copystat_swap_fixed_strict.sh` — negative-assertion wrappers over the reproducer. They **pass** only when the swap fails to flip the target's mode.
- [x] `tests/regression/t01/test_copystat_swap_with_no_copy_stat.sh` — confirms `-n` / `--no-copy-stat` still disables the metadata copy entirely.
- [x] `tests/regression/t01/test_copystat_positive.sh`, `test_copystat_various_modes.sh`, `test_copystat_mode_mask.sh` — positive regressions: with no attacker, the output file's mode is copied from the input (including across a range of modes and the `S_IRWXU|S_IRWXG|S_IRWXO` mask).
- [x] `tests/regression/t01/test_copystat_timestamp.sh` — confirms atime/mtime are propagated via `futimens` on the still-open fd.
- [x] `tests/regression/t01/test_copystat_stdout_skip.sh` — exercises the `current_output_path == NULL` branch (`echo ... | brotli -c | brotli -dc`); no metadata call is attempted on stdout.
- [x] `tests/regression/t01/test_copystat_stdin_input.sh` — confirms stdin input (no input path to stat from) is handled cleanly.
- [x] `tests/regression/t01/test_copystat_roundtrip.sh` — plain compress/decompress roundtrip still produces byte-identical output, so the reorder in `CloseFiles()` has not regressed the happy path.
- [x] `tests/regression/t01/run_all.sh` — runs all of the above and summarizes pass / fail / infra-skip counts.
- [x] Built and re-ran the full set under ASan + UBSan (`build-asan/`); no sanitizer reports.

To reproduce locally:

```sh
cc -fPIC -shared -o tests/regression/t01/libfclose_swap.so \
    tests/regression/t01/fclose_swap.c -ldl
tests/regression/t01/run_all.sh \
    build-asan/brotli \
    tests/regression/t01/libfclose_swap.so
```

The `.so` and `plain.bin.br` artifacts are intentionally not committed (covered by the existing `*.so` / `*.br` ignore rules) — they are built on demand from `fclose_swap.c` and `plain.bin`.

The companion path-TOCTOU on `OpenOutputFile()` is tracked separately in google/brotli#1462 and will land in its own PR, to keep the `-f` behavior change out of this review.


eustas:
Hello. Thanks for your PR. Changes to `brotli.c` looks good, but tests should be reworked.
`plain.bin` should be generated due to its redundant nature. Tests should become a suite rather than bunch of scripts. There is no nice "unit test" solution for bash, so please consider converting those scripts to Python.

parasol-aser:
Addressed the test feedback:
- replaced the T-01 bash scripts with one Python unittest suite;
- removed checked-in plain.bin/target fixtures; the suite generates them in temp dirs;
- kept the fclose swap helper as source and builds it on demand for the symlink-swap regression cases.

Validation: python3 -m py_compile tests/regression/t01/copystat_regression_test.py; ASAN_OPTIONS=detect_leaks=0 python3 tests/regression/t01/copystat_regression_test.py ./build-asan/brotli -v.

eustas:
Looks good. You just need to sign CLA before landing can happen.

parasol-aser:
> Looks good. You just need to sign CLA before landing can happen.

thanks! signed.

eustas (approved):

PR #1474, opened 2026-05-04
Avoid signed overflow in compound dictionary decoding

ZX41R:
Follow-up to 4792c8e.

`EnsureCompoundDictionaryInitialized` advances an `int cursor` by
`1 << block_bits`. For large compound dictionaries that are still accepted by
`AttachCompoundDictionary`, `block_bits` can be 23 and `cursor` reaches
2139095040 before the next increment overflows signed int. Keeping this cursor
as `size_t` makes the loop arithmetic match the non-negative size domain it is
walking.

`InitializeCompoundDictionaryCopy` had the same issue in its `address + length`
bounds check. This changes that check to use `BROTLI_SAFE_ADD` before comparing
the result with `total_size`.

Local checks:
- rebuilt `brotlidec` with ASan/UBSan
- the compound-dictionary reproducer that previously hit the signed overflow
  now returns `_SUCCESS`
- the same reproducer with `-fwrapv` no longer trips ASan
- `ctest --output-on-failure`: 28/28 passed

### DIFF
--- a/c/dec/decode.c
+++ b/c/dec/decode.c
@@ -1556,16 +1556,16 @@ static void EnsureCompoundDictionaryInitialized(BrotliDecoderState* state) {
   BrotliDecoderCompoundDictionary* addon = state->compound_dictionary;
   /* 256 = (1 << 8) slots in block map. */
   int block_bits = 8;
-  int cursor = 0;
+  size_t cursor = 0;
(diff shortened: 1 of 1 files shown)


eustas:
Should be covered by #1479

PR #1494, opened 2026-06-06
fix(java): reject decoder push length exceeding input buffer capacity

aleister1102:
### Summary
`org.brotli.wrapper.dec.DecoderJNI.Wrapper.push(int length)` only rejected `length < 0`. The native input buffer is allocated once in `nativeCreate` with a fixed capacity, and `push(length)` reports how many bytes are available in that existing buffer — but nothing checked `length` against the buffer capacity, and the native side did not retain the allocation size. An oversized `length` was forwarded directly into `BrotliDecoderDecompressStream`.

### Reproduction (low-level public API)
```java
DecoderJNI.Wrapper decoder = new DecoderJNI.Wrapper(1); // capacity = 1
decoder.getInputBuffer().put(0, (byte) 0);
decoder.push(32);   // claims 32 bytes available in a 1-byte buffer
```
The decoder then receives `(in = 1-byte allocation, in_size = 32)` and reads past the allocation — a heap out-of-bounds read (observed crash in `DecodeVarLenUint8` under a guard allocator). The high-level `Decoder.decode()` / `Decoder.decompress()` paths are unaffected because they always pass `length <= capacity`.

### Fix
- `DecoderJNI.Wrapper.push()` now also rejects `length > inputBuffer.capacity()`, completing the existing input validation. No API or behavior change for legitimate callers (whose `length` is already `<= capacity`).
- Defense in depth: `DecoderHandle` now stores the allocation size (`input_size`), and `nativePush()` rejects `input_length > input_size` (returning the existing ERROR status) before mutating decoder state.
- Added `DecoderJNITest`: oversized `push` is rejected, negative `push` is rejected, and a valid `push(capacity)` round trip still decodes correctly.

### Testing
```sh
cd java
bazel test //org/brotli/wrapper/dec/...
```
All decoder wrapper tests pass (`DecoderJNITest`, `DecoderTest`, `CornerCasesTest`, `BrotliInputStreamTest`, `BrotliDecoderChannelTest`).

Note: the regression test covers the public `Wrapper.push()` guard. The native `input_size` check is defense in depth and is not reachable through the public API once the Java guard is in place.

Issue number: https://issuetracker.google.com/issues/520744758

### DIFF
--- a/java/org/brotli/wrapper/dec/BUILD.bazel
+++ b/java/org/brotli/wrapper/dec/BUILD.bazel
@@ -104,3 +104,16 @@ java_test(
     test_class = "org.brotli.wrapper.dec.CornerCasesTest",
     runtime_deps = [":test_lib"],
 )
+
+java_test(
+    name = "DecoderJNITest",
+    size = "large",
+    data = [
+        ":brotli_jni",  # Bazel JNI workaround
+    ],
+    jvm_flags = [
+        "-DBROTLI_JNI_LIBRARY=$(location :brotli_jni)",
+    ],
+    test_class = "org.brotli.wrapper.dec.DecoderJNITest",
+    runtime_deps = [":test_lib"],
+)
--- a/java/org/brotli/wrapper/dec/DecoderJNI.java
+++ b/java/org/brotli/wrapper/dec/DecoderJNI.java
@@ -58,6 +58,9 @@ public class DecoderJNI {
       if (length < 0) {
         throw new IllegalArgumentException("negative block length");
       }
+      if (length > inputBuffer.capacity()) {
+        throw new IllegalArgumentException("block length exceeds input buffer capacity");
+      }
       if (context[0] == 0) {
         throw new IllegalStateException("brotli decoder is already destroyed");
       }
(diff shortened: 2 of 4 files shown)


aleister1102:
@eustas, friendly ping when you have a chance - would you be able to review and approve the workflow run (first-time contributor)?

This is a small, self-contained hardening fix: `DecoderJNI.Wrapper.push()` now rejects an input length larger than the allocated input buffer, with a defense-in-depth check in `nativePush`, preventing an out-of-bounds read reachable through the public low-level wrapper API. It adds a regression test, and the high-level `Decoder` paths are unchanged. All checks (including the CLA) are green.

Thanks!

eustas:
`DecoderJNI` should have had `default` visibility (package-private) and thus no unsanitized parameters would ever be passed to Wrapper.
But alas, it is public, and thus part of API...

Let's make check in `DecoderJNI.Wrapper` and not in native; that is the main purpose of it, anyways.

aleister1102:
> `DecoderJNI` should have had `default` visibility (package-private) and thus no unsanitized parameters would ever be passed to Wrapper. But alas, it is public, and thus part of API...
> 
> Let's make check in `DecoderJNI.Wrapper` and not in native; that is the main purpose of it, anyways.

Makes sense - thanks. The check now lives only in `DecoderJNI.Wrapper.push()`; the native guard has been removed. Tests still pass through the Wrapper API.

eustas (approved):

PR #1498, opened 2026-06-25
Fix jint->size_t integer narrowing in DecoderJNI and EncoderJNI nativePush

momo-123-coder:
Fix jint→size_t integer narrowing in DecoderJNI and EncoderJNI nativePush

### Summary

`nativePush` in both `decoder_jni.cc` and `encoder_jni.cc` accepts
`jint input_length` (a signed 32-bit JNI integer) and assigns it directly
to a `size_t` field without a sign check:

    handle->input_length = input_length;   // decoder_jni.cc:133
    handle->input_last   = input_length;   // encoder_jni.cc:147

When the Java caller passes a negative value (e.g., -1, or any value
where the high bit is set), the implicit jint → size_t conversion wraps
to SIZE_MAX - |value| + 1 (4 294 967 295 for -1 on 32-bit, or
18 446 744 073 709 551 615 on 64-bit). This causes:

- in_size = handle->input_length - handle->input_offset → huge value
- BrotliDecoderDecompressStream is called with a multi-gigabyte declared
  input size pointing at attacker-controlled memory
- On instrumented builds: heap-buffer-overflow / process abort
- On production: out-of-bounds read, potential information disclosure,
  or remote code execution depending on heap layout

### Root cause

`jint` is defined as `int32_t` in <jni.h>. The JNI specification allows
any Java `int` value, including negative values, to reach native code.
There is no existing guard against `input_length < 0` before the assignment.

### Fix

Add a sign check and replace the implicit narrowing cast with an explicit
static_cast<size_t> in both files:

    if (input_length < 0) {
      return;
    }
    handle->input_offset = 0;
    handle->input_length = static_cast<size_t>(input_length);

The encoder_jni.cc path is fixed identically (input_last field).

### Testing

Existing unit tests continue to pass. A LibFuzzer harness that exercises
this path with negative input_length values is available on request.

### Security impact

- CWE-191: Integer Underflow (Wrap or Wraparound)
- CWE-125: Out-of-Bounds Read
- CVSS 3.1: AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:N/A:H — 9.1 CRITICAL
  (network-reachable through any JVM processing untrusted compressed data)

### DIFF
--- a/java/org/brotli/wrapper/dec/decoder_jni.cc
+++ b/java/org/brotli/wrapper/dec/decoder_jni.cc
@@ -129,8 +129,12 @@ JNIEXPORT void JNICALL Java_org_brotli_wrapper_dec_DecoderJNI_nativePush(
     if (handle->input_offset < handle->input_length) {
       return;
     }
+    /* Reject negative sizes: jint is signed; implicit cast to size_t wraps. */
+    if (input_length < 0) {
+      return;
+    }
     handle->input_offset = 0;
-    handle->input_length = input_length;
+    handle->input_length = static_cast<size_t>(input_length);
   }
 
   /* Actual decompression. */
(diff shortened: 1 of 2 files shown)


eustas:
JNI interface is private exactly for that reason: parameters are sanitized before being passed to native.

PR #1518, opened 2026-08-09
Do not read ISUNCOMPRESSED after a metadata meta-block

mohass1927:
`decodeMetaBlockLength` reads the `ISUNCOMPRESSED` bit after a metadata meta-block, which the format does not contain. RFC 7932 §9.2 places that bit only in the non-metadata branch, and the C decoder agrees: `BROTLI_STATE_METABLOCK_HEADER_METADATA` increments `meta_block_remaining_len` and returns, so it never reaches `BROTLI_STATE_METABLOCK_HEADER_UNCOMPRESSED`.

The Java metadata branch returns early only when `MSKIPBYTES == 0`. With `MSKIPBYTES >= 1` it falls through to the header tail shared with the non-metadata branch and reads the bit.

Both branches then run `if (isMetadata || isUncompressed) jumpToByteBoundary()`, which is why this is usually invisible — when the metadata header ends unaligned, the jump absorbs the extra bit and both decoders land on the same byte. It stops being invisible when the header ends *exactly* on a byte boundary: C jumps zero bits, Java has already consumed one and jumps seven more, so Java is a full byte ahead for the rest of the stream and reads every later meta-block header from a different offset.

Because a meta-block header starts wherever the previous one ended, that alignment is reachable, and a stream can be arranged so the subsequent headers are individually valid under *both* offsets. Then the two decoders produce different output from the same bytes and both report success rather than one of them erroring.

### What this changes

`ISUNCOMPRESSED` is now read inside the non-metadata branch of `decodeMetaBlockLength`, so the metadata branch cannot reach it. No exit point is added; `s.metaBlockLength++` stays common to both branches, matching `++s->meta_block_remaining_len` in the C metadata state.

`csharp/org/brotli/dec/Decode.cs` gets the identical change. `java/org/brotli/dec/kt/Decode.kt`, `go/brotli/decode.go`, `js/decode.ts` and `js/decode.js` are left to the transpiler.

### Test

`SynthTest.testMetadataEndingOnByteBoundary`, a 14-byte stream whose metadata header ends at bit 96. The layout is confirmed by `research/brotlidump.py`:

```
bit  0      WBITS: window 16
bit  1-81   compressed meta-block: MLEN 4, simple literal code {"A", "B"}, emits "ABAB"
bit 82-95   metadata header: ISLAST 0, MNIBBLES "11" (metadata), reserved 0,
            MSKIPBYTES 1, MSKIPLEN byte 00
bit 96      header ends, 96 % 8 == 0
byte 12     the skipped metadata byte
byte 13     final empty meta-block
```

The existing metadata coverage in `testPeculiarWrap` does not catch this: it uses `MSKIPBYTES == 1`, so it does reach the offending line, but its header ends unaligned so the jump hides the bit.

### Verified

`javac` + JUnit 4.13.2, OpenJDK 21, on the same tree with only `Decode.java` swapped between `master` and this change:

| | without the fix | with the fix |
|---|---|---|
| `SynthTest` | `Tests run: 47, Failures: 1` — `testMetadataEndingOnByteBoundary`, `BrotliRuntimeException: Error code: -4` | `OK (47 tests)` |
| `BitReaderTest` / `DecodeTest` / `DictionaryTest` / `EagerStreamTest` / `CompoundDictionaryTest` / `TransformTest` | OK (2 / 8 / 1 / 1 / 3 / 3) | unchanged |

C# (`dotnet build brotlidec.csproj` + `dotnet test brotlidec.Tests.csproj`, SDK 10.0.303): `Passed! - Failed: 0, Passed: 46`. On the same 14 bytes, a console harness over the C# sources prints `FAIL IOException: Brotli stream decoding failed` with `Decode.cs` at `master` and `OK len=4 text=ABAB` with this change.

Reference C decoder on the same 14 bytes: 4 bytes, `ABAB`.


eustas:
Nice catch! Lets polish.
 - drop changes in `.kt`, `.go`, `.js`, `.ts`: those will be overwritten by transpiler anyways
 - but leave changes in `.cs` (transpilation is currently dusty)
 - please check that test fails without fix
 - instead of adding exit point, lets pull `if (s.inputEnd == 0) ...` into `else` branch;

mohass1927:
Thanks — all four done, PR is now three files: `java/org/brotli/dec/Decode.java`, `java/org/brotli/dec/SynthTest.java`, `csharp/org/brotli/dec/Decode.cs`.

**Dropped `.kt`, `.go`, `.js`, `.ts`.** Also dropped the Go mirror of the test (`go/brotli/synth_test.go`), since that file is transpiled too.

**Kept `.cs`.** Same shape as the Java change. `dotnet build brotlidec.csproj` + `dotnet test brotlidec.Tests.csproj` (SDK 10.0.303) → `Passed! - Failed: 0, Passed: 46`. Decoding the new 14-byte vector through the C# sources: `FAIL IOException: Brotli stream decoding failed` with `Decode.cs` at `master`, `OK len=4 text=ABAB` with the change.

**No extra exit point.** `if (s.inputEnd == 0) ...` moved into the `else` branch. `s.metaBlockLength++` stays common to both branches, which also matches `++s->meta_block_remaining_len` in the C metadata state:

```java
    } else {
      for (int i = 0; i < sizeNibbles; ++i) {
        ...
      }
      /* ISUNCOMPRESSED is present only in non-metadata meta-blocks; see RFC 7932 section 9.2. */
      if (s.inputEnd == 0) {
        s.isUncompressed = BitReader.readFewBits(s, 1);
      }
    }
    s.metaBlockLength++;
    return BROTLI_OK;
```

**Test fails without the fix.** Same tree, `javac` + JUnit 4.13.2 on OpenJDK 21, only `Decode.java` swapped for the `master` copy:

```
master Decode.java
  .........................................E......
  There was 1 failure:
  1) testMetadataEndingOnByteBoundary(org.brotli.dec.SynthTest)
  java.lang.AssertionError: expected to succeed decoding, but failed
  Caused by: org.brotli.dec.BrotliRuntimeException: Error code: -4
      at org.brotli.dec.Decode.readComplexHuffmanCode(Decode.java:634)
  Tests run: 47,  Failures: 1

patched Decode.java
  ...............................................
  OK (47 tests)
```

`BitReaderTest`, `DecodeTest`, `DictionaryTest`, `EagerStreamTest`, `CompoundDictionaryTest` and `TransformTest` pass identically either way (2 / 8 / 1 / 1 / 3 / 3 tests).

One thing worth flagging before you look at CI: `//org/brotli/dec:SynthTest_kt` runs this same `SynthTest` against `org.brotli.dec.kt.BrotliInputStream`, so with the `.kt` hunk dropped the new test is red on the java bazel job until the transpiler runs. Measured with kotlinc 2.1.20 against the patched `SynthTest`:

```
kt/Decode.kt at master  ->  Tests run: 47, Failures: 1
                            org.brotli.dec.kt.BrotliRuntimeException: Error code: -4
+ the one-line hunk     ->  OK (47 tests)
```

Happy either way — say the word and I will put the `.kt` hunk back, or leave it to the transpiler.

eustas (approved):

eustas:
Hi. Let's then update `.kt` (for easier landing), and later it will be rewritten (or not) by transpiler.

PR #1520, opened 2026-08-12
enc: reject max_base64_regions greater than 16384

adilburaksen:
## Problem

`BROTLI_ALLOC` (`c/enc/memory.h`) computes `N * sizeof(T)` with no overflow
check, and `BROTLI_PARAM_MAX_BASE64_REGIONS` accepted any `uint32_t`. Setting
it close to `SIZE_MAX / sizeof(Base64Region)` makes the `base64_regions`
allocation in `HasherSetup` (`c/enc/hash.h`) wrap, producing a buffer far
smaller than the `num_base64_regions < params->max_base64_regions` guards in
`c/enc/backward_references_inc.h` assume.

## Fix

Validate the parameter where it is set: cap it at 16384 and reject larger
values, in the same style as `BROTLI_PARAM_STREAM_OFFSET`.

## Testing

32-bit ASan build (`gcc -m32 -fsanitize=address`, so `sizeof(size_t) == 4` and
`sizeof(Base64Region) == 8`), input containing a `;base64,` region:

| `MAX_BASE64_REGIONS` | before | after |
| --- | --- | --- |
| `1u << 29` | `SetParameter` -> 1, then ASan heap-buffer-overflow (WRITE of size 4, `backward_references_inc.h:77` in `CreateBackwardReferencesNH5`) | `SetParameter` -> 0, compresses cleanly |
| `UINT32_MAX` | accepted | rejected |
| 16385 | accepted | rejected |
| 16384 | accepted | accepted, compresses cleanly |
| 16 (default) | accepted | accepted, byte-identical output |

## Reachability

`BROTLI_PARAM_BASE64_MODE` defaults to disabled, and neither new parameter is
exposed by the bundled `brotli` CLI or by the Python/Java/Go bindings in this
repository, so this is a public C API contract fix for the recently added
Base64 detection feature (037b70e), not a live issue in any current consumer.


eustas:
Let's set a limit to  be `16384`; instead of clamping let's reject setting greater values. No need for lengthy comments: just validate parameter when it is set.

adilburaksen:
Done — moved the check to `BrotliEncoderSetParameter`: values above 16384 are now rejected with `BROTLI_FALSE` instead of clamped, and `c/enc/hash.h` is back to its original form with no added comments. The limit is exposed as `BROTLI_MAX_BASE64_REGIONS` next to the existing default, and the parameter's doc comment notes it. Rebased on current master.

Re-checked under the same 32-bit ASan build: `1u << 29` and `UINT32_MAX` are rejected and compress cleanly afterwards, 16384 and the default 16 still work. Happy to move the constant out of the public header if you'd rather keep the literal local to `encode.c`.

eustas (approved):

eustas:
Need to update encode.h.3 as well... I'll check how to do it properly (ideally we should have a script + CI to check it on GitHub side).

PR #1524, opened 2026-08-22

3412-cloud:

## Problem

`EncoderJNI.Wrapper.push(Operation, int length)` only rejected negative
lengths. A length larger than `inputBuffer.capacity()` was passed to
`nativePush`, which forwarded it as `available_in` to
`BrotliEncoderCompressStream`. The encoder then read up to `length` bytes
past the end of the input buffer (heap out-of-bounds read, up to the value
of `length`).

Reported as https://issuetracker.google.com/issues/549278294 (heap OOB read
in brotli Java JNI encoder).

## Fix

Mirror the decoder-side fix (commit 2dc2a5f, "fix(java): reject decoder
push length exceeding input buffer capacity"):

1. `EncoderJNI.Wrapper.push` now rejects `length > inputBuffer.capacity()`
   with `IllegalArgumentException` (same message style as the decoder).
2. Defense in depth in the native push path: `EncoderHandle` now records
   the allocated `input_size` and `nativePush` returns early when
   `input_length` is negative or exceeds it.

## Test

`EncoderJNITest` covers:
- push length exceeding the input buffer capacity is rejected;
- negative length is rejected;
- a push of exactly the buffer capacity round-trips successfully.

## Notes

- The C API contract already requires `available_in` to not exceed the
  buffer; the JNI wrapper is the layer responsible for enforcing it for
  Java callers.
- This is the encoder-side sibling of the previously fixed decoder bug;
  the encoder path was not covered by the original fix.


3412-cloud:
Hi maintainers — this fix has been open a few days without review. It adds the missing upper-bound check in the Java encoder path (EncoderJNI.Wrapper.push): a push length exceeding the input buffer capacity is now rejected instead of triggering an out-of-bounds read (reported upstream, Google issue 549278294). +98/-0 with tests included. @eustas would you mind taking a look when you have a moment? Happy to adjust anything.

eustas:
There is a comment for Decoder counterpart (https://github.com/google/brotli/pull/1494#issuecomment-4682575325):
```
DecoderJNI should have had default visibility (package-private) and thus no unsanitized parameters would ever be passed to Wrapper.
```
Unlike `DecoderJNI` here access is package-private, so only fair users are in control of "length".

I could only agree that `EncoderJNI.java` changes are somewhat useful (more to fix the contract, than actually check).

PR #1528, opened 2026-08-31
cli: preserve sub-second mtime on the BSDs (fixes #1515)

### LINKED ISSUE #1515: mtime's nsec are truncated on OpenBSD and FreeBSD (at least)
Hello,

#891 and #992 did some changes to the time-copy functions, but OpenBSD and FreeBSD (at a minimum) look like they were left out.

```
### OpenBSD, on FFS
# gzip -fk -9 obsd.html
# brotli -kf9 obsd.html
# zstd -fkT0 -19 obsd.html
obsd.html            : 52.08%   (   891 B =>    464 B, obsd.html.zst)          
# stat -f "%Fm %z %N" obsd.html*
1785531532.061542134 891 obsd.html
1785531532.000000000 387 obsd.html.br
1785531532.061542134 461 obsd.html.gz
1785531532.061542134 464 obsd.html.zst
# brotli -V
brotli 1.2.0
```

cf. https://github.com/openbsd/ports/tree/master/archivers/brotli

```
### FreeBSD 14.4 and 15.0, on ZFS
# man 1 brotli > brotlitest
# brotli brotlitest
# zstd -fkT0 brotlitest
brotlitest           : 40.65%   (  5.60 KiB =>   2.28 KiB, brotlitest.zst)
# stat -f '%Fm %N' brotlitest*     
1785531998.940700000 brotlitest
1785531998.000000000 brotlitest.br
1785531998.000000000 brotlitest.zst
# brotli -V
brotli 1.2.0
```

cf. https://github.com/freebsd/freebsd-ports/tree/main/archivers/brotli

* Affected: OpenBSD, FreeBSD, maybe more? I don't have any other machines to test this on!
* Unaffected: Linux (Archlinux)

tvost2:
### Fixes #1515

`HAVE_UTIMENSAT` in `c/tools/brotli.c` is only set to 1 when
`_POSIX_C_SOURCE >= 200809L` or `_ATFILE_SOURCE` is defined:

```c
#if defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 200809L)
#define HAVE_UTIMENSAT 1
#elif defined(_ATFILE_SOURCE)
#define HAVE_UTIMENSAT 1
#else
#define HAVE_UTIMENSAT 0
#endif
```

FreeBSD, OpenBSD, NetBSD and DragonFly BSD ship `utimensat()` / `futimens()`
and nanosecond-resolution `struct stat` timestamps, but do **not** define
either feature macro by default. So on those systems the CLI takes the
`utime()` fallback and the copied mtime is truncated to whole seconds — as
reported in #1515, where the `.br` output lands on `…​.000000000` next to a
source with real sub-second precision (`.gz` / `.zst` keep it).

`#891` / `#992` added the nanosecond path but the platform ladder never got a
BSD branch.

### Change

Add an explicit `elif` for the four BSDs. The existing non-Apple
`ATIME_NSEC` / `MTIME_NSEC` macros already use the `st_atim` / `st_mtim`
spelling those systems expose, so nothing else changes. Purely additive —
Linux / macOS / Windows are untouched.

### Testing

The preprocessor selection is verified with a extracted copy of the ladder:
`-D__FreeBSD__`, `-D__OpenBSD__`, `-D__NetBSD__`, `-D__DragonFly__` each yield
`HAVE_UTIMENSAT=1`; the bare case (no POSIX/ATFILE macro) still yields `0`.
CLI still builds clean (clang, `cmake -GNinja`).

I don't have a BSD host to check the round-trip end to end; the change is a
one-line platform-detection addition mirroring the existing branches, and the
BSD `futimens` / `st_mtim` APIs it relies on are long-standing
(FreeBSD ≥ 11, OpenBSD ≥ 5.0, NetBSD ≥ 6.0).

---

AI-assisted; analysed, changed and build-checked by me, and I take
responsibility for it.

### DIFF
--- a/c/tools/brotli.c
+++ b/c/tools/brotli.c
@@ -79,6 +79,9 @@ static int ms_open(const char* filename, int oflag, int pmode) {
 #define HAVE_UTIMENSAT 1
 #elif defined(_ATFILE_SOURCE)
 #define HAVE_UTIMENSAT 1
+#elif defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__) || \
+    defined(__DragonFly__)
+#define HAVE_UTIMENSAT 1
 #else
 #define HAVE_UTIMENSAT 0
 #endif


eustas:
SGTM, but please drop comment. If one ever wonders why we have this `elif` this PR with explanation will be found.

tvost2:
Done — dropped the comment, force-pushed. The rationale lives in the commit message / this PR now. Thanks for the quick look!

eustas (approved):

PR #1537, opened 2026-09-06
CLI: report a failed output write and keep the input

lenamonj:
When writing the output failed, `brotli --rm` still deleted the input and exited 0. `CloseFiles` never flushed `fout` itself: the only flush was inside `CopyStat`, whose failure returns silently, and the unlink of the input ran regardless of `is_ok`.

`CloseFiles` now flushes first and reports a failure, skips the metadata copy after one, and unlinks the input only when the output is known complete. `CopyStat` no longer flushes.

`brotli --rm -f -o /dev/full f` on master: exit 0, `f` gone. With this change: exit 1, `f` kept.


lenamonj:
@googlebot I signed it!

eustas:
Comment in `CopyStat` is a bit confusing (because we enter it if there were no failure) and not necessary (explains not doing something we no longer do here). Please drop it. Thanks.

lenamonj:
Comment dropped.

PR #1538, opened 2026-09-06
CLI: refuse an output path that names the input file

lenamonj:
`brotli -f -o F F` (and `-d -f -o F F`) opened the output with `O_TRUNC` before reading the input, so `F` was emptied and the tool then compressed nothing, or decompressed nothing and unlinked it.

`OpenFiles` now compares the two paths by `st_dev`/`st_ino` before opening the output and refuses when they name one file. An inode of 0, which `stat()` reports on Windows, is treated as unknown and the tool proceeds as before.


lenamonj:
@googlebot I signed it!

eustas (approved):

PR #1539, opened 2026-09-06
Call the lazy static initializer from EnsureStaticInit

lenamonj:
In `BROTLI_STATIC_INIT_LAZY` mode nothing called the embedder's `BrotliXxxLazyStaticInit`: `BrotliXxxEnsureStaticInit` returned `kLazyInitOk`, which only `BrotliXxxLazyStaticInitInner` sets, so every encoder and decoder operation failed. The decoder's one-shot `BrotliDecoderDecompress` never checked it at all, since the check lived in `BrotliDecoderCreateInstance`.

`EnsureStaticInit` now calls `LazyStaticInit` before reading the flag, matching the header's description of the mode, and the decoder's check moves into `BrotliDecoderStateInit`, which every state passes through. `dec/static_init_lazy.cc` mirrors the encoder's reference implementation.

Built with `-DBROTLI_STATIC_INIT=2` and the two `.cc` files, a compress/decompress roundtrip fails on master and passes with this change; the default-mode CMake tests are unchanged.


lenamonj:
@googlebot I signed it!

eustas (dismissed):

eustas:
Does `c/dec/decode.c` still need `static_init.h` include (since invocation is moved to `state`)?
Also, in state invoke `ensure` before `decoder start` callback, so that possible initialization cost is not accounted towards decoder instance (since it is shared).

lenamonj:
Removed the include; nothing in decode.c referenced static_init.h once the call moved. BrotliDecoderEnsureStaticInit now runs before BROTLI_DECODER_ON_START, so the initialisation cost stays outside the instance callback. 73 of 73 tests pass.

